PR #8, opened 2022-05-31
Add code to handle LIMIT 0 queries to return names and types 

genzgd:


clickhouse_connect/driver/httpclient.py
```diff
@@ -112,7 +115,8 @@ def _apply_settings(self, settings: Dict[str, Any] = None):
                 self.common_settings[key] = str(value)
 
     def _format_query(self, query: str) -> str:
-        if query.upper().strip().startswith('INSERT ') and 'VALUES' in query.upper():
+        query = query.strip()
+        if query.upper().startswith('INSERT ') and 'VALUES' in query.upper():
             return query
         if not query.strip().endswith(self.read_format):
```
guykoh, line 121: nit: you can replace now 'query.strip()' with query

clickhouse_connect/driver/httpclient.py
```diff
@@ -127,8 +131,22 @@ def exec_query(self, query: str, settings: Optional[Dict] = None, use_none: bool
...
             params.update(settings)
-        response = self._raw_request(self._format_query(query), params=params, headers=headers)
-        data_result = self.parse_response(response.content, use_none)
+        columns_only = columns_only_re.search(query)
+        if columns_only:
+            final_query = query + ' FORMAT JSON'
+        else:
+            final_query = self._format_query(query)
+        response = self._raw_request(final_query, params=params, headers=headers)
+        if columns_only:
```
guykoh, line 140: nit: I would combine the 2 if cases it will lead to duplicate 

response = self._raw_request(final_query, params=params, headers=headers)

but overall will be shorter and more readable IMO

clickhouse_connect/driver/httpclient.py
```diff
@@ -127,8 +131,22 @@ def exec_query(self, query: str, settings: Optional[Dict] = None, use_none: bool
...
+            json_result = json.loads(response.content)
+            names: List[str] = []
+            types: List[ClickHouseType] = []
+            for col in json_result['meta']:
```
guykoh, line 144: Can you please add a comment with the json format structure?

guykoh (approved):

PR #9, opened 2022-06-01
Removing sql comments from query before removing all new lines.

guykoh:
Reason: In line comments in sql are declared using '--' identifier. Once we remove all new lines from the command, inline comments will comment out all the trailing sql commands after the comment line. This is why it is important to remove all inline comments before removing all the new lines.


genzgd:
How do you feel about just leaving in the newlines?  I debated internally at one point and didn't consider comments, and at this point I'm not sure it's worth the edge cases to try to parse and remove comments with regexes

PR #15, opened 2022-06-15
Fix Encoding issue (Chinese character cause request to fail)

1c7:
For example:
UnicodeEncodeError: 'latin-1' codec can't encode character '\u7a7a' in position 57: Body ('空') is not valid Latin-1. Use body.encode('utf-8') if you want to send it encoded in UTF-8.

### DIFF
--- a/clickhouse_connect/driver/httpclient.py
+++ b/clickhouse_connect/driver/httpclient.py
@@ -205,7 +205,7 @@ class HttpClient(Client):
             response: Response = self.session.request(method, self.url,
                                                       headers=headers,
                                                       timeout=(self.connect_timeout, self.read_timeout),
-                                                      data=data,
+                                                      data=data.encode(),
                                                       params=params)
         except RequestException as ex:
             logger.exception('Unexpected Http Driver Exception')


1c7:
# Context
* https://github.com/dbt-labs/dbt-core/issues/5371
* https://github.com/ClickHouse/dbt-clickhouse/issues/57

1c7:
## Here is my test code to make sure this PR indeed fixed ths issue   

```python
import clickhouse_connect

  host='ec2-xx-131-xxx-146.cn-north-1.compute.amazonaws.com.cn', 
  port=8123, 
  username='default',
  password="[hide]",
  database='mp_production',
)

query_result = client.query("""
  SELECT id FROM apply_answers WHERE value_string NOT IN ('空') LIMIT 10
""")

print (query_result.result_set) 
```

`空` just mean "empty"

genzgd:
Thanks for reporting this issue and submitting a pull request!  I've implemented a slightly different fix because data may not always be a string in the raw_request method.  This fix will be part of the 0.1.2 release

1c7:
Got it, Thanks!


https://github.com/ClickHouse/clickhouse-connect/commit/cd7b0f164aca98187009f5a9f2ca97b7736db5d7
image

1c7:
by the way, please check out this: https://github.com/dbt-labs/dbt-core/issues/5375#issuecomment-1156109300

PR #20, opened 2022-06-17
Don't try to send readonly settings (even for send_progress or compression)

genzgd:


clickhouse_connect/driver/httpclient.py
```diff
@@ -107,15 +106,19 @@ def __init__(self, interface: str, host: str, port: int, username: str, password
         self.connect_timeout = connect_timeout
         self.read_timeout = send_receive_timeout
         if send_progress:
-            self.session.params['send_progress_in_http_headers'] = '1'
-            self.session.params['wait_end_of_query'] = '1'
+            settings['send_progress_in_http_headers'] = '1'
```
guykoh, line 109: Are you changing the settings that you received as an argument?
I don't think this is smart. Maybe create a shallow copy at least?

genzgd: It's probably not necessary since the **kwargs dictionary is a new mapping built as part of the method call, but it does look weird to modify what looks like a mutable argument

guykoh (approved):

PR #21, opened 2022-06-26
Add socket options for TCP keep alive to HTTP adapter

genzgd:
I believe this fixes the timeout issues reported in dbt-clickhouse.  Silly me for thinking that a requests keep alive connection would also add keep alive to the TCP socket.  See the discussion below 

https://www.finbourne.com/blog/the-mysterious-hanging-client-tcp-keep-alives
https://github.com/psf/requests/issues/3353

### DIFF
--- a/clickhouse_connect/driver/client.py
+++ b/clickhouse_connect/driver/client.py
@@ -19,6 +19,7 @@ class Client(metaclass=ABCMeta):
     Base ClickHouse Connect client
     """
     column_inserts = False
+    valid_transport_settings = set()
 
     def __init__(self, database: str, query_limit: int, uri: str, settings: Dict[str, Any] = None):
         """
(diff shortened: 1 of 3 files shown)


clickhouse_connect/driver/client.py
```diff
@@ -47,7 +47,7 @@ def _apply_settings(self, settings: Dict[str, Any] = None):
     def _validate_settings(self, settings: Dict[str, Any]):
         validated = {}
         for key, value in settings.items():
-            if 'session' not in key:
+            if 'session' not in key and key not in ('database', 'buffer_size'):
```
guykoh, line 50: So 'database', 'buffer_size' and keys with 'session' could be missing or readonly and then we set them regardless?
Maybe check that they are not readonly? Or at least add a comment explaining why it is safe to override those readonly settings?

genzgd: database and buffer_size are not actual settings, they can always be sent as query parameters with any HTTP request, so they aren't ever "read only".  I wanted to make sure they weren't stripped from the request by the logic that checks valid settings.

genzgd: Same with the session parameters "session_id" and "session_timeout"

guykoh (approved):

PR #26, opened 2022-07-12
Include pip installation

mharrisb1:
Since the README is serving as the only real source of documentation for the project you should include the pip installation instructions and optionally a link to PyPi.

I had members of my team run into issues with the `setup.py` installation because they didn't even know the project was available through PyPi. This is the easiest route for installation and is standard to include in documentation for Python projects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,10 +6,15 @@ ClickHouse HTTP interface.
 
 
 ### Installation
+
+```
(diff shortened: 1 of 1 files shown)


genzgd (approved):

genzgd:
Thanks!  I completely agree.

PR #27, opened 2022-07-28
fix future warnings pyarrow api

pindash:


genzgd:
Hi -- thanks for your PR!  The arrow implementation is actually going to change in the next release (which should be in the next few days) to return a PyArrow Table instead of the current tuple of schema plus record batch, since it currently returns incomplete data for large result sets.  Your improvement will be included in that release.

If you don't mind, let me know if the PyArrow Table approach works for your use case.

pindash:
thanks genzgd, that works!

PR #33, opened 2022-08-12
Arrow inserts

genzgd:
-- Always backtick quote identifiers in SQLAlchemy and client insert statements
-- Clean up push workflow to not require pip install -e (this makes tests and lint checking possible without setting an explicit version in setup.py, which we only do on release
-- Minor refactoring for inserts to cleanly support direct inserts of arrow tables

### DIFF
--- a/clickhouse_connect/driver/query.py
+++ b/clickhouse_connect/driver/query.py
@@ -167,15 +167,20 @@ local_tz = datetime.now().astimezone().tzinfo
 BS = '\\'
 must_escape = (BS, '\'')
 
+
 def quote_identifier(identifier: str):
-    if identifier.startswith('`'):
+    first_char = identifier[0]
+    if first_char in ('`', '"') and identifier[-1] == first_char:
+        # Identifier is already quoted, assume that it's valid
         return identifier
     return f'`{identifier}`'
+
 
 def finalize_query(query: str, parameters: Optional[Dict[str, Any]], server_tz: Optional[tzinfo] = None) -> str:
     if not parameters:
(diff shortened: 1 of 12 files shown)


clickhouse_connect/driver/query.py
```diff
@@ -167,13 +167,16 @@ class DataResult(NamedTuple):
 BS = '\\'
 must_escape = (BS, '\'')
 
+def quote_identifier(identifier: str):
+    if identifier.startswith('`'):
```
guykoh, line 171: what about '"'?

tests/test_requirements.txt
```diff
@@ -1,3 +1,5 @@
+requests
+pytz
```
guykoh, line 2: Where do you use it?

guykoh (approved):

clickhouse_connect/driver/query.py
genzgd: Good catch, updated (and added a test).

tests/test_requirements.txt
genzgd: requests is used in clickhouse_connect/driver/httpclient and pytz is used in a couple of places, include clickhouse_connect/datatypes/temporal.  They used to get pulled in to the test environment as I side effect of installing clickhouse_connect using pip (pip install -e .), but pip didn't like the "developer_only" version. 

So instead I don't install the project at all (which actually is cleaner) and install all the test and lint requirements explicitly (as well as compiling the cython modules explicitly as well).

PR #51, opened 2022-10-06
Remove forced `enable_http_compression=1` on the HTTP request level

slvrtrn:
Do not enforce `enable_http_compression=1` setting if `HttpClient(compress)` is set to `True` - it is now required to have it either in the server's `users.xml` or supply it separately on the query level.


genzgd:
What is the reason for doing this?

PR #52, opened 2022-10-07
Add Python code highlighting

peter279k:
# Changed log

- Making these Python codes highlighting.


genzgd:
Thanks for the update.  The README is due for an update in the next couple of weeks anyway, as most of the documentation is moving to the main ClickHouse site here:  https://clickhouse.com/docs/en/integrations/language-clients/python/intro.

PR #71, opened 2022-11-10
Use CaseInsensitiveDict for sqla_type_map

EugeneTorap:
Hi @genzgd
We use ClickHouse in apache/superset. Before `clickhouse-connect` we used `clickhouse-sqlalchemy` driver.
When we migrated to `clickhouse-connect` we noticed that ClickHouse type name is changed from UPPER to Pascal case.
All our superset datasets columns have UPPER case types in superset DB and we have the next error:

> error: "Unrecognized ClickHouse type base: UINT64 name: UINT64"

We need to write a flexible logic to ignore case of type name.
I prepare `CaseInsensitiveDict` class in order to make all item keys lowercase and get it also by lowercase key.
This PR resolves our problem of migrating from `clickhouse-sqlalchemy` to `clickhouse-connect`

### DIFF
--- a/clickhouse_connect/cc_sqlalchemy/datatypes/base.py
+++ b/clickhouse_connect/cc_sqlalchemy/datatypes/base.py
@@ -115,7 +115,7 @@ class CaseInsensitiveDict(dict):
         return super().__getitem__(item.lower())
 
 
-sqla_type_map: CaseInsensitiveDict[str, Type[ChSqlaType]] = CaseInsensitiveDict()
+sqla_type_map: Dict[str, Type[ChSqlaType]] = CaseInsensitiveDict()
 schema_types = []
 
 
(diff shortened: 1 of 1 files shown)


EugeneTorap:
@genzgd Thank you!
I've created a https://github.com/apache/superset/pull/22039 PR for apache/superset in order to use `clickhouse-connect` by default.
Let me know when you release/push 4.0.1 version to PyPI

genzgd:
Wow, thanks for submitting a Superset PR -- we have that on our todo list
but hadn't quite gotten there.

I have a 0.4.1 branch in progress and hope to release it in the next day or
two.  Do you have time to migrate the tests as Ville requested or do you
need help with that?

EugeneTorap:
I will try to migrate the tests myself

PR #87, opened 2022-12-29
File insert, server side bindings

genzgd:


.github/workflows/on_push.yml
```diff
@@ -50,7 +50,8 @@ jobs:
         clickhouse-version:
           - '22.3'
           - '22.8'
-          - '22.9'
+          - '22.10'
+          - '22.11'
```
mshustov, line 54: + 22.12?

CHANGELOG.md
```diff
@@ -7,6 +7,19 @@ other client methods, `get_client` now accepts an optional `settings` Dict[str,
 to set ClickHouse server settings.  The use of `**kwargs` for this purpose is deprecated and will be removed in a future
 release.
 
+## 0.4.8, 2022-12-29
+### New Features
+* File upload support.  The new `driver.tools` module adds the function `insert_file` to simplify
```
mshustov, line 12: So many awesome features! Don't forget to update the official docs https://clickhouse.com/docs/en/integrations/language-clients/python/driver-api

.github/workflows/on_push.yml
genzgd: well that was `latest` when I started this PR :)

PR #92, opened 2023-01-09
Streaming, C improvements, refactoring

genzgd:


clickhouse_connect/driver/transform.py
```diff
@@ -1,38 +1,116 @@
...
+                except (StopIteration, IndexError):
+                    source.close()
+                    return None
+                result_block = []
+                for col_num in range(num_cols):
+                    name = source.read_leb128_str()
+                    if block_num == 0:
+                        names.append(name)
+                    type_name = source.read_leb128_str()
+                    if block_num == 0:
```
slvrtrn, line 36: is the second if statement necessary? seems to be the same as on the line 33

clickhouse_connect/driver/transform.py
```diff
@@ -1,38 +1,116 @@
...
+                    yield next_block
+
+        return QueryResult(None, gen(), tuple(names), tuple(col_types), context.column_oriented, source)
+
+    @staticmethod
+    def build_insert(context: InsertContext):
+        if context.compression == 'gzip':
+            compressor = GzipCompressor()
+        else:
+            compressor = NullCompressor()
```
slvrtrn, line 66: do we need to create new instances here every time? If we have a single thread anyway, probably it's safe to flush Zlib from a static object

tests/integration_tests/test_streaming.py
```diff
@@ -0,0 +1,13 @@
...
+
+def test_numbers_stream(test_client: Client):
+    query_result = test_client.query('SELECT number FROM numbers(1000000) LIMIT 1000000', column_oriented=True)
+    total = 0
+    blocks = 0
+    with query_result:
+        for x in query_result.stream_blocks():
+            total += len(x[0])
+            blocks += 1
+    assert blocks > 0
```
slvrtrn, line 12: does it vary?

clickhouse_connect/driverc/buffer.pyx
```diff
@@ -0,0 +1,223 @@
...
+        return column
+
+    def read_array(self, t: str, unsigned long long num_rows):
+        cdef array.array template = array_templates[t]
+        cdef array.array result = array.clone(template, num_rows, 0)
+        cdef unsigned long long sz = result.itemsize * num_rows
+        if self._set_slice(sz) == 255:
+            raise StopIteration
+        memcpy(result.data.as_voidptr, self._cur_slice(), sz)
+        if must_swap:
```
slvrtrn, line 180: does it work without `global must_swap`? If Cython is alike normal Python from what I recall, this might be `false` all the time

clickhouse_connect/driver/transform.py
```diff
@@ -1,38 +1,116 @@
...
+
+
+class GzipCompressor:
+    def __init__(self, level: int = 6, wbits: int = 31):
+        self.zlib_obj = zlib.compressobj(level=level, wbits=wbits)
+
+    def compress_block(self, block):
+        return self.zlib_obj.compress(block)
+
+    def complete(self):
```
jkaflik, line 115: not a strong opinion, but it makes more sense as `flush` or `close`

clickhouse_connect/driver/transform.py
jkaflik: `context.compression` could be pass with an actual compressor instance

clickhouse_connect/driver/httputil.py
```diff
@@ -0,0 +1,84 @@
...
+from requests.adapters import HTTPAdapter
+from urllib3 import poolmanager
+
+
+# Increase this number just to be safe when ClickHouse is returning progress headers
+http._MAXHEADERS = 10000  # pylint: disable=protected-access
+
+KEEP_INTERVAL = 30
+KEEP_COUNT = 3
+KEEP_IDLE = 30
```
jkaflik, line 17: I guess these are default values. If yes, would it make sense to have it with `DEFAULT_` prefix?

clickhouse_connect/driver/buffer.py
```diff
@@ -0,0 +1,119 @@
...
+must_swap = sys.byteorder == 'big'
+
+
+class ResponseBuffer(ByteSource):
+    slots = 'slice_sz', 'buf_loc', 'end', 'gen', 'buffer', 'slice'
+
+    def __init__(self, source):
+        self.slice_sz = 4096
+        self.buf_loc = 0
+        self.end = 0
```
jkaflik, line 15: it's a buffer size?

clickhouse_connect/driver/buffer.py
```diff
@@ -0,0 +1,119 @@
...
+        self.source = source
+        self.gen = source.gen
+        self.buffer = bytes()
+
+    def read_bytes(self, sz: int):
+        if self.buf_loc + sz <= self.end:
+            self.buf_loc += sz
+            return self.buffer[self.buf_loc - sz: self.buf_loc]
+        bridge = bytearray(self.buffer[self.buf_loc: self.end])
```
jkaflik, line 24: not sure I understand the meaning of a `bridge`. it's a remaining buffer?

clickhouse_connect/driver/transform.py
genzgd: Going with Serge here, we only need one instance. :)

clickhouse_connect/driver/buffer.py
genzgd: I like that name better -- originally it wasn't actually a buffer size, but the end of the "good" data in a fixed buffer, but the implementation changed.

PR #95, opened 2023-01-13
Compression updates, performance improvements

genzgd:


clickhouse_connect/driver/compression.py
```diff
@@ -0,0 +1,79 @@
...
+
+try:
+    import brotli
+except ImportError:
+    brotli = None
+
+
+def available_compression():
+    comp = ['lz4', 'zstd']
+    if brotli:
```
jkaflik, line 17: basically, brotli will be not returned to the list until the package is available, that's correct?

PR #135, opened 2023-02-22
add bug report template

mshustov:
## Summary
Adds a bug report template


mshustov:
it's been added already. closing

PR #140, opened 2023-03-04
Allows connect to server version 19.16.19.85

aastafiev:
Hello,

This PR brings backward compatibility with ClickHouse version 19.16.19.85.

This query won't work because readonly field doesn't exit.
`SELECT name, value, readonly FROM system.settings LIMIT 10000`
Consequently, clickhouse-connect can't connect to the server.

I suggest including common setting "readonly" to be able set it in runtime and change query to 
`SELECT name, value FROM system.settings LIMIT 10000`
for server version less than 22.3.

So, client could set "readonly" setting before connecting to the server like so:
```
import clickhouse_connect
from clickhouse_connect import common

common.set_setting('readonly', 1)
client = clickhouse_connect.get_client(host='localhost', user='default', port=9123)
```


genzgd:
Thanks for your submission!  A couple of general comments --

- If possible, please start your branch from a more recent version so there aren't so many merge commits
- In your IDE, please set your line wrapping to 120 characters so not so many existing lines get reformatted

As for your main change, it seems fine to allow backward compatibility.  I think you could simplify the code quite a bit though by keeping just one query for system.settings, something like this:

```python
if self.min_version('22.3'):   # We should check this version, when was the readonly flag introduced?
     read_only = 'readonly'
else:
     read_only = common.get_setting('readonly')
server_settings = self.query(f'SELECT name, value, {readonly} as readonly FROM system.settings LIMIT 10000')
```

playtest.py
```diff
@@ -1,19 +1,37 @@
+from io import BytesIO
```
genzgd, line 1: Could you add a separate test module for your testing?  This file is just a small example of how to connect to play.clickhouse.com

clickhouse_connect/common.py
```diff
@@ -59,3 +60,4 @@ def _init_common(name: str, options: Sequence[Any], default: Any):
 _init_common('dict_parameter_format', ('json', 'map'), 'json')
 _init_common('invalid_setting_action', ('send', 'drop', 'error'), 'error')
 _init_common('product_name', (), '')
+_init_common('readonly', (0, 1, 2), 0)
```
genzgd, line 63: I think we could limit this to 0 and 1 since it's only necessary for older versions

.gitignore
```diff
@@ -15,10 +16,13 @@
...
+.env
 
 # Python cruft
 *.pyc
 .python-version
 
 # C extensions
 *.pyd
+
+playtest_custom.py
```
genzgd, line 28: You can put your own test files under the `dev` directory without a special case in .gitignore

genzgd (changes requested):

playtest.py
aastafiev: fixed tests for servers version started from 19.16.19.85

clickhouse_connect/common.py
aastafiev: done

.gitignore
aastafiev: done

aastafiev:
Hello,

Thank you for your feedback.
I've taken into account your commets and passed through all tests starting server version 19.16.19.85.

playtest.py
aastafiev: revert back original file

genzgd:
Thanks for backporting all those tests!  That's awesome.  We can't merge this until you sign the CLA though.  Also if you could remove the flake8 comments (I see a couple of `# noqa comments`) that would reduce the noise a bit.

aastafiev:
Will do it asap. 
I don’t understand how to sign CLA. Could you give me instructions on it?


---------------------
Sincerely yours,
Aleksey Astafiev

aastafiev:
I've signed the CLA but for some reason aastafiev marked as signed but Aleksey Astafyev not. Is it critical?

genzgd (approved):

aastafiev:
Thank you for acceptence.
Could you reveal when to expect a new release with this feature?

genzgd:
Is it something you need available for install soon?  We can do a release if you need it, but otherwise I would probably wait until there's a new feature or important bug fix before cutting another release, since I can't imagine many folks are running ClickHouse 19.x :)

aastafiev:
If it wouldn't take more than month I could wait :)

genzgd:
It's never that long :).   If there's not anything else to add in the next week or so I'll cut a release.

PR #147, opened 2023-03-13
fix key error in connection with no available settings from server

Barsoomx:
## Summary
If connection settings aren't available/exposed, any connection fails after initial settings select
```sql
SELECT name, value, readonly as readonly FROM system.settings
```

image

## Checklist
not relevant


genzgd (approved):
Thanks!  What permissions caused this to happen?

PR #161, opened 2023-04-03
Fixes #159 - quoting and escaping string literals inside lists

### LINKED ISSUE #159: Array(String) query parameters don't quote the strings
### Describe the bug

When using an `Array(String)` in server-side params the strings inside the array are sent unquoted, so the query crashes

### Steps to reproduce
1. `client.query('SELECT {l:Array(String)}', parameters={"l": ["a"]}).result_rows`
2. > Cannot parse quoted string: expected opening quote ''', got 'a'. (CANNOT_PARSE_QUOTED_STRING) (version 22.8.15.23 (official build))

### Expected behaviour

Returns `['a']`

### Code example
```python
import clickhouse_connect
    dsn=connection_string
)
client.query('SELECT {l:Array(String)}', parameters={"l": ["a"]}).result_rows
```

```
Code: 26. DB::ParsingException: Cannot parse quoted string: expected opening quote ''', got 'a'. (CANNOT_PARSE_QUOTED_STRING) (version 22.8.15.23 (official build))
```

#### Environment
* clickhouse-connect version:  `clickhouse-connect==0.5.16`
* Python version: 3.7
* Operating system: Ubuntu

#### ClickHouse server
* ClickHouse Server version:22.8.15.23

### Additional info

It really looks like `format_bind_value` doesn't quote Strings, because when we use a `String` at the top level we send `param_x="something"` and not `param_x="'something'"`, but this logic only holds true at the top level.
I think `format_bind_value` should have an optional `top_level=True` parameter, when it calls itself recursively, it should call itself with `top_level=False` and when the input is a string, if `not top_level` it should return `'{input}'` (though with a bit of sanitazing to avoid injections)

ewjoachim:
## Summary
Fixes #159 - quoting and escaping string literals inside lists

## Note

The PR could be completed with more tests (e.g. tuples, dicts, etc) and more scoped tests (e.g. unit testing `format_str` and `escape_str` themselves). Feel free to take over the PR and do it.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ in a future release.  Starting with 0.5.9 the driver now requests ClickHouse pro
 The secondary effect of the `send_progress` argument -- to set `wait_end_of_query=1` -- is now handled automatically based
 on whether the query is streaming or not.
 
+## [next release]
+### Fix quoting and escaping of array literals in server parameters
+See [#159](https://github.com/ClickHouse/clickhouse-connect/issues/159)
+
 ## 0.5.18, 2023-03-30
 ### Performance Improvement
 - The server timezone will not be applied (and Python datetime types will be timezone naive) if the client and server timezones match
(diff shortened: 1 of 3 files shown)


genzgd (approved):

genzgd:
This is great, thanks so much!

PR #173, opened 2023-04-25
nit: change unexpected http error log to warning level

alexandrosandre:
## Summary

In this error scenario the library is logging an error and raising it at the same time. This contradicts the generally adopted behavior of logging debug messages only and raising error scenarios as exceptions to the caller. As a result, callers cannot decide on their own whether they want to handle the exception silently (e.g. with their own retries) or not, because their log will always include the ERROR log of the library.

Not sure whether we should delete the log, I went for the less aggressive change in case anyone created dependencies on it.

## Checklist


genzgd (approved):

genzgd:
Yeah, exception seems like overkill :)

alexandrosandre:
LMK if this needs a bullet point in CHANGELOG.

alexandrosandre:
Done, but not sure about your policy.

genzgd (approved):

PR #185, opened 2023-05-02
V0.5.23

genzgd:
## Summary
Fixes for several recently reported minor bugs

## Checklist


clickhouse_connect/driver/client.py
```diff
@@ -436,21 +437,27 @@ def query_arrow(self,
...
         :param settings: Optional dictionary of ClickHouse settings (key/string values)
         :param use_strings:  Convert ClickHouse String type to Arrow string type (instead of binary)
+        :param external_data ClickHouse "external data" to send with query
         :return: PyArrow.Table
         """
         settings = dict_copy(settings)
         if self.database:
             settings['database'] = self.database
-        if arrow_str_setting in self.server_settings and arrow_str_setting not in settings:
+        if self._setting_status(arrow_str_setting).is_writable and arrow_str_setting not in settings:
```
cpcloud, line 454: Does `use_strings=True` silently have no effect if the setting isn't writable?

It seems like the fact that the setting is tied to global user permissions at all is very strange.

I understand the design of clickhouse's configuration permissions system is out of scope for this PR. Should I raise this on the main clickhouse issue tracker?

tests/integration_tests/test_external_data.py
```diff
@@ -16,6 +19,20 @@ def test_external_simple(test_client: Client):
     assert result[0][0] == '12 Angry Men'
 
 
+def test_external_arrow(test_client: Client):
+    if not arrow:
+        pytest.skip('PyArrow package not available')
```
cpcloud, line 24: It might make sense to use `pytest.importorskip("pyarrow")` to avoid the boilerplate of having to write an error message.

clickhouse_connect/driver/client.py
genzgd: Currently it's silent, but I will add a log warning if it's explicitly set and still not writable.

The workaround for the permissions system is for the system administrator to set `readonly=2` for the setting in question.  See some discussion here:  https://clickhouse.com/docs/en/operations/settings/constraints-on-settings.  

I don't think that's likely to change as we strongly enforce the idea that "the administrator said you can't change it, even for this query", since there are so many settings and some of them can almost certainly be misused.

play.clickhouse.com is of course a special case because it's open to everyone, so that's always going to be extremely locked down.

genzgd: The other option is to fail the query (more gracefully), but that seems like overkill?

genzgd: I decided to throw an OperationalException, since that seems easier to debug that tracking down a warning.

cpcloud: I suppose it's a bit off topic, but I'll just log my protest here regarding the options-require-permissions design 😃 

This design decision makes it extremely cumbersome to build libraries that rely on ClickHouse. I cannot depend on being able to set any option even if that option is totally benign like being able to configure some minute detail about the result format.

It seems like session-user-scoped options would be ideal here. I guess these exist (sort of?) as passing `SETTINGS` per-query, but again are totally defeated by the strange permissions model.

genzgd: Well this is definitely a ClickHouse server project level issue.  Maybe the simplest solution is to default query settings to `readonly=2` (the readonly=2 option is actually fairly recent).  I understand your frustration (and it makes implementing an HTTP client that is heavily dependent on certain settings quite quickly), so maybe it's work opening an issue (if there's not one already) in the main ClickHouse project for discussion.

cpcloud (approved):

cpcloud:
@genzgd Thanks for you work on this, very much appreciated.

PR #197, opened 2023-06-07
fix _MAXHEADERS property

istrebitel-1:
## Summary
Connection pool uses http.client.._MAXHEADERS property, not http._MAXHEADERS

You cat try in on by example:
```python
import http

http._MAXHEADERS = 1  # not valuable setting
http.client._MAXHEADERS = 1  # valuable setting

client_config = {
    'host': '127.0.0.1',
    'port': 8123,
    'username': 'dev',
    'password': 'dev',
}

client = get_client(
    **client_config
)

client.query("""
    SELECT sleepEachRow(1), *
    FROM numbers(10)
    SETTINGS receive_timeout = 99990, send_timeout = 99990
""")
```


genzgd (approved):

genzgd:
Thanks for the fix! -- did you discover the problem because something broke?

istrebitel-1:
> Thanks for the fix! -- did you discover the problem because something broke?

Yes, it was large query and its cancelled by python in 3h 15min, but in clickhouse query was still running

genzgd:
Thanks, good to know moving forward.  This fix should get in the next minor release, probably next week.

PR #214, opened 2023-07-03
refactor: pkg_resources -> importlib.metadata

cwegener:
pkg_resources is in the process of being deprecated in favor of
importlib.metadata and importlib.resources

importlib.metadata has been available in the standard library since 3.8

And as of 3.10 the importlib.metadata API is no longer provisional

In order to support Python version older than 3.10, the
importlib_metadata backport is used instead of the actual
importlib.metadata API.

(NOTE: There was no explicit install_require entry that needed removing
in setup.py for the setuptools package which provides the pkg_resources
module. It seems that it was merely assumed that setuptools is always
present, which is actually not the case.)

### DIFF
--- a/clickhouse_connect/common.py
+++ b/clickhouse_connect/common.py
@@ -2,15 +2,15 @@ import sys
 from dataclasses import dataclass
 from typing import Any, Sequence, Optional, Dict
 
-import pkg_resources
+from importlib_metadata import PackageNotFoundError, distribution
 
 from clickhouse_connect.driver.exceptions import ProgrammingError
 
 
 def version():
     try:
-        return pkg_resources.get_distribution('clickhouse-connect').version
-    except pkg_resources.ResolutionError:
+        return distribution('clickhouse-connect').version
+    except PackageNotFoundError:
         return 'development'
 
 
--- a/clickhouse_connect/entry_points.py
+++ b/clickhouse_connect/entry_points.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 
 import sys
-import pkg_resources
+from importlib_metadata import PackageNotFoundError, distribution
 
 EXPECTED_EPS = {'sqlalchemy.dialects:clickhousedb',
                 'sqlalchemy.dialects:clickhousedb.connect'}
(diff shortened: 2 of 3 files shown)


genzgd:
Thanks for your submission!  To be honest really hate to add another third party dependency even if it's tiny -- is there a way to make that conditional on the Python version?  

Maybe the better way to do this is to just read the existing version file some way instead of relying on the pkg_resources at all?  That probably a bit more work but seems safer -- I can switch to that approach instead in the next release.  I assume that your goal is to avoid deprecation warnings on install?

cwegener:
>  To be honest really hate to add another third party dependency

This change is not adding a third party dependency.

`importlib_metadata` is the official backport of `import.metadata` from the standard library.

Since `clickhouse-connect` supports (the now end of life) Python 3.7, the backport is required in order to use `importlib.metadata` functionality.

> is there a way to make that conditional on the Python version?

Yes, but I don't think it's worth the additional code.

> Maybe the better way to do this is to just read the existing version file some way

I agree (at leas for the `clickhouse.connect:version()` case, For the "Entrypoints validation utility", I actually was not able to find out what the purpose that that utility is)

>  I assume that your goal is to avoid deprecation warnings on install?

No, that was not my goal.
The goal is to avoid the side effects of using `pkg_resources` in general, which results in a full evaluation of the `setup.py` and all the associated version constraint check that setuptools will run on the `install_requires` list.  For packaging reasons, I cannot strictly comply with all the version constraint checks that are specified in the Clickhouse Connect `setup.py` and I must rely on the package distribution versions that are provided by the Operating System, which are slightly out of bounds of the version constraints in `setup.py`.

But yes, deprecations must be addressed regardless.

cwegener:
> > To be honest really hate to add another third party dependency
> 
> This change is not adding a third party dependency.
> 
> `importlib_metadata` is the official backport of `import.metadata` from the standard library.

Technically speaking it would be called 'third party' in a sense that is outside of the standard library though.

genzgd:
Ah, got it, thanks for the detailed response!   (The entrypoints utility is really for developer use since debugging whether entrypoints are correctly created is annoying.)

I'd be happy to merge this as is for now if you wouldn't mind signing the CLA and if I get a chance I'll see if I can remove the dependency entirely.

cwegener:
> I'd be happy to merge this as is for now if you wouldn't mind signing the CLA
Done.

I did just see that I didn't touch the tests that are using the `pkg_resources`. I can have a look at that in a separate change.

genzgd:
If you don't get to it I'll see if I can fix the tests.  I completely forgot about those.

Issue #239, opened 2023-09-15
Empty Map with nested Map Types raises an IndexError.


CaptainCuddleCube:
### Describe the bug

If your query returns an empty map, but the map is of a next Map type, an IndexError is raised. 
The query works fine on the server.

### Steps to reproduce
1. `client.query("select Cast(([],[]), 'Map(String, Map(String, String))')")`

### Expected behaviour
It should not raise an error, and return an empty dictionary.

### Code example
```python
import clickhouse_connect
dsn='http://admin:dev@localhost:8123/default'
client = clickhouse_connect.get_client(dsn=dsn)
client.query("select Cast(([],[]), 'Map(String, Map(String, String))')")
# IndexError: array index out of range
```

#### Environment
* clickhouse-connect version: 0.6.11
* Python version: 3.11.5
* Operating system: Linux/MacOS

#### ClickHouse server
* ClickHouse Server version: 23.6.2.18
* ClickHouse Server non-default settings, if any: N/A
* `CREATE TABLE` statements for tables involved: N/A

CaptainCuddleCube:
It's a one line fix in the source code, just update to the following line of code:
in file `clickhouse_connect/datatypes/containt.py:183`
```
        total_rows = 0 if len(offset) == 0 else offsets[-1]
```

genzgd:
Fixed in 0.6.12

PR #240, opened 2023-09-15
fix: Empty nests raising out of IndexError.

CaptainCuddleCube:
## Summary
Fixes this bug: https://github.com/ClickHouse/clickhouse-connect/issues/239

When you have an empty map that has a type of nests maps, and IndexError is raise; this fixes that issue. 

Example: 
```python
client.query("select Cast(([],[]), 'Map(String, Map(String, String))')")
```

## Checklist

### DIFF
--- a/clickhouse_connect/datatypes/container.py
+++ b/clickhouse_connect/datatypes/container.py
@@ -180,7 +180,7 @@ class Map(ClickHouseType):
     # pylint: disable=too-many-locals
     def read_column_data(self, source: ByteSource, num_rows: int, ctx: QueryContext):
         offsets = source.read_array('Q', num_rows)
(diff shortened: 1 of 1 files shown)


CaptainCuddleCube:
I'm not really sure where to add the test to verify this issue.

genzgd:
Thanks for the contribution!  There probably should be some reorganization of the integration test files and this kind of test issue should go into something like `test_datatypes.py`.  If you want to put it in `tests/integration/test_native.py` for the moment that would be fine, and if you don't get to it I can add something later.

CaptainCuddleCube:
I'm struggling to get the tests working, but it's really easy to test: just create an empty map with the type `Map(String, Map(String, String))`.

PR #288, opened 2024-01-16
Fixes setup query params

aastafiev:
## Summary
In our case clickhouse http/https interface lies behind additional "access http layer" (AHL). The base client class initialization (`Client`) needs to run couple sql requests. But they could't succeed without going through AHL. So, we made minor changes to continue using this great library.

After this PR we could add additional options to `valid_transport_settings` attribute of `HttpClient` before running queries.

## Checklist
- [v] Unit and integration tests passed


genzgd:
If I understand your use case correctly, you need to add some custom query parameters to every request, but Client initialization failed because those parameters were not recognized as valid settings.  It seems like you could change the [global setting](https://clickhouse.com/docs/en/integrations/python#global-settings) for `invalid_setting_action` from `error` to `send` and add the settings without a code change (although this will generate logger warnings).

Alternatively, because `valid_transport_settings` is a class variable, you could just add your new parameter names to the class before initializing a Client like:

```python
HttpClient.valid_transport_settings.add('my_query_param_name')
```

which again would not require a code change.   This PR seems to remove some of the guard rails around sending bad settings, unless I'm not understanding something?

aastafiev:
Unfortunatly, it doesn't work, because when we set additional parameter in `valid_transport_settings` it should be applyed for the first queries to clickhouse, for example
`self.server_version, server_tz = \
            tuple(self.command('SELECT version(), timezone()', use_database=False))`
But on this stage `self.params` is not initialized (it will be initialized only after `super().__init__(..)` in `HttpClient`) and `_raw_request` goes without nessasary transport settings in http query params (`final_params = dict_copy(self.params, params)`: here `self.params` is empty and `params` wasn't set before by definition). So, we unable to pass transport layer.

On the initial stage we define clickhouse server_settings for the client, but before we need to access it. I just added ability to take in consideration transport setting from the first step of client initializing. I don't think it conflicts with "some of the guard rails around sending bad settings", cause we doesn't send any bad server settings at this time of initialization. This server setting would by critical on user quering stage.

genzgd:
Ah, I understand, and now that I think about it I believe other users have had similar issues with the initial ClickHouse queries when the HttpClient is initialized.  I think my preferred approach would be to add something like a `preinit` hook in the HttpClient `__init__` method, where the user could specify a function that could be used to modify the client object before the initial queries were run (including updating the `params` and `valid_transport_settings` fields).  Alternatively adding a class level dictionary of "global" query parameters to the class object that would automatically added to the `ch_settings` dictionary in the `__init__` method might be a more specific solution to your issue.

Also at the moment, only "new" query parameters ("settings") are validated on each request, so I think the code here which bypasses the "server_settings" check could result in invalid settings being passed at query time, which would likely cause query failures that are easier to debug when the Client is initiated.

If you want to modify the PR according to one of the approaches above I'm happy to take another look; otherwise if you just want to open an enhancement/bug issue along the lines of "allow sending custom query parameters with all requests" I would probably get to it in a week or two."

aastafiev:
I can do it.
I think the second suggest than given ability a user to pass a function for changing internal state of client class. It doesn't sounds robust

genzgd:
sounds good, thanks!

aastafiev:
Well :)
Better late than never.
I realised that HttpClient already has `params` class attribute of type dict. So, using it in cooperation with `valid_transport_settings` solve our case.
 
This PR could be rejected.

genzgd:
Ah, that does work for the initial queries, but you also have to send it as a setting.  In the next release I'll make it so the class level `params` are carried over into the instance level `params` to make it a little cleaner.

aastafiev:
Don’t worry I’ll do it. 


---------------------
Sincerely yours,
Aleksey Astafiev

PR #291, opened 2024-01-18
Pass init params

aastafiev:
## Summary
Initial HttpClient parameters passing to query settings
Added possibility to disable http request timeouts

I couldn't find better solution to test passing parameter that put SQL_test setting before client creation in integration tests.


genzgd (approved):

genzgd:
Thanks!  I'll try to get a new release out in the next few days.

PR #301, opened 2024-02-13
Correct typing of Client.query*(query)

biggerfisch:
## Summary

Despite a type hint of `str`, the default parameter for these methods is `None`. This makes type analyzers sad.

To fix this, hint the type as `Optional[str]` to match reality.

It's not correct to say that `query` is a required, non-None type due to the possibility of a QueryContext with an existing query. Therefore, this is the more correct change.

## Checklist


genzgd:
Thanks for the cleanup!

Issue #316, opened 2024-03-13
Make X-ClickHouse-Summary available through cursor API


elchyn-cheliabiyeu:
We want to get summary information from the query we run using the cursor object. This field is not a specified cursor field, it only relates to clickhouse

elchyn-cheliabiyeu:
I need an approval and will make MR for this

genzgd:
I'm not sure what this would look like but it seems useful.  We're always happy to review community pull requests, but we do ask that you include tests for the new functionality.

PR #317, opened 2024-03-14
Add summary field to the Cursor object

elchyn-cheliabiyeu:
## Summary

Add summary field to Cursor object to retrieve the result of 'X-Clickhouse-Summary' header
Fixes https://github.com/ClickHouse/clickhouse-connect/issues/316

## Checklist

### DIFF
--- a/clickhouse_connect/dbapi/cursor.py
+++ b/clickhouse_connect/dbapi/cursor.py
@@ -1,7 +1,7 @@
 import logging
 import re
 
-from typing import Optional, Sequence
+from typing import Optional, Sequence, List, Dict
 
 from clickhouse_connect.datatypes.registry import get_from_name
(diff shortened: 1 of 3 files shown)


clickhouse_connect/dbapi/cursor.py
```diff
@@ -29,6 +30,7 @@ def __init__(self, client: Client):
         self.names = []
         self.types = []
         self._rowcount = 0
+        self._summary: List[Dict[str, str]] = []
```
genzgd, line 33: It's been a while since I looked at this, but can the cursor be reused?  Is that why you want to keep a list of the summaries instead of just the latest?

elchyn-cheliabiyeu: Actually it is because of executemany

genzgd (approved):

PR #328, opened 2024-03-28
fix(external-data): allow empty data external data

martijnthe:
### Summary

Before this patch, the `ExternalData` / `ExternalFile` APIs would not create an empty table when `data=b''` was being passed, because `data` would be checked for truthiness. This PR fixes this bug by using `data is not None` checks instead.

 ### Test Plan

Added an integration test to cover this case.

## Checklist


genzgd:
Thanks for the fix with test!  The number of edge cases folks find to break continues to surprise me.

PR #331, opened 2024-04-01
chore(typing): improve type annotations of Client.raw_query

martijnthe:
## Summary

The return type of the `raw_query` method of a `Client` instance depends on the `stream: bool` argument: if it's `True`, the return type is an `io.IOBase` and if it's `False`, the return type is `bytes`.  This PR adds `@overload`s to make type checkers (such as mypy and pyright) aware of this, to avoid needing to to type-cast at the call-site.

 ## Test Plan

I started a `tests/typing/` folder for type testing and added a `client_typing.py` file with all possible type variations of the `stream` argument.

I manually ran `pyright tests/typing/client_typing.py` and observed there were no errors. Ideally, running a type checker would be part of this project's CI set up, but this is out of scope for this PR.

## Checklist

### DIFF
--- a/tests/typing/client_typing.py
+++ b/tests/typing/client_typing.py
@@ -1,0 +1,21 @@
+import io
+
+from clickhouse_connect.driver import Client
+from typing import assert_type, cast, Union
+
+# Test Client.raw_query overloads:
+
+client = cast(Client, None)
+
+result = client.raw_query("")
+assert_type(result, bytes)
+
+result = client.raw_query("", stream=False)
+assert_type(result, bytes)
+
+result = client.raw_query("", stream=True)
+assert_type(result, io.IOBase)
+
+stream = cast(bool, None)
+result = client.raw_query("", stream=stream)
+assert_type(result, Union[bytes, io.IOBase])
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ ClickHouse Connect has been included as an official Apache Superset database con
 However, if you need compatibility with older versions of Superset, you may need clickhouse-connect
 v0.5.25, which dynamically loads the EngineSpec from the clickhouse-connect project.
 
+## UNRELEASED
+### Improvements
+- Added type `@overload`s for the `raw_query` method of `Client` for correct inference of the return
+  type based on the `stream: bool` argument. Thanks for [Martijn Thé](https://github.com/martijnthe) for the PR!
+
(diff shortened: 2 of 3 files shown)


tests/typing/client_typing.py
```diff
@@ -0,0 +1,21 @@
...
+
+result = client.raw_query("", stream=False)
+assert_type(result, bytes)
+
+result = client.raw_query("", stream=True)
+assert_type(result, io.IOBase)
+
+stream = cast(bool, None)
+result = client.raw_query("", stream=stream)
+assert_type(result, Union[bytes, io.IOBase])
```
martijnthe, line 21: See PR description:

> I manually ran `pyright tests/typing/client_typing.py` and observed there were no errors. Ideally, running a type checker would be part of this project's CI set up, but this is out of scope for this PR.

Would you be open to a PR that adds a CI job to run [pyright](https://github.com/microsoft/pyright) on the code base?

genzgd: I don't see any reason not to?  So I think that a PR for that makes sense.

genzgd: In this particular case maybe some refactoring makes sense to separate a `stream` response (which is only used by arrow) from a regular bytes response, instead of making it an overload.  The overload approach seems to add a lot of repetition.  I'll give it some more thought this week.

martijnthe: > In this particular case maybe some refactoring makes sense to separate a stream response (which is only used by arrow) from a regular bytes response, instead of making it an overload.

Ah, I guess you are considering the “stream” parameter to not be part of the public API then?

genzgd: At this point no, I'd rather have a wrapper or some other method of signaling that the method returns a stream vs a bunch of bytes instead of an optional parameter that changes the return type.  I didn't really give it a lot of thought when considering the community PR that made the change.

genzgd:
So I split `raw_query` into `raw_query` and `raw_stream` -- I don't know if you still want to add the `pyright` functionality but at this point I think enough has changed that a fresh PR is easier.  We continue to appreciate your efforts in keeping the typing in good shape.

martijnthe:
> So I split `raw_query` into `raw_query` and `raw_stream` -- I don't know if you still want to add the `pyright` functionality but at this point I think enough has changed that a fresh PR is easier. We continue to appreciate your efforts in keeping the typing in good shape.

Ok thanks! The typing on the public API looks good 👍

PR #338, opened 2024-04-08
apply_server_timezone - condition change

KrzysztofPrzygodzki:
## Summary
`apply_server_timezone` was evaluated to `FALSE`, because first condition was checking if `apply_server_timezone` is equal to `"always"`. However, when you initiate class `HttpClient`, that inherits form `Client` class, you set `apply_server_timezone` to `True`. 

Second condition (line 68 after `or`) was evaluated to `False` in my case. The function `coerce_bool` returns `True`, because you check if `apply_server_timezone` is `True`, but  `offset_differ` is evaluated in my system as `False`. My local TZ == server TZ and you check for the difference.

Please look on the example produced by the current code:

```Python
client.command("SELECT serverTimeZone()")
'Europe/Warsaw'
```

For 2024-04-01 result is as expected.

```Python
client.command("SELECT toDateTime('2024-04-01')")
2024-04-01 00:00:00'
```
```Python
client.query("SELECT toDateTime('2024-04-01')").result_rows
[(datetime.datetime(2024, 4, 1, 0, 0, tzinfo=datetime.timezone(datetime.timedelta(seconds=7200), 'CEST')),)]
```

But for 2024-02-01 `command` is correct, but `query` is wrong - it returns `2024, 2, 1, 1, 0` instead of `2024, 2, 1, 0, 0`:
```Python
client.command("SELECT toDateTime('2024-02-01')")
2024-02-01 00:00:00'
```
```Python
client.query("SELECT toDateTime('2024-02-01')").result_rows
[(datetime.datetime(2024, 2, 1, 1, 0, tzinfo=datetime.timezone(datetime.timedelta(seconds=7200), 'CEST')),)]

### DIFF
--- a/clickhouse_connect/driver/client.py
+++ b/clickhouse_connect/driver/client.py
@@ -64,7 +64,7 @@ class Client(ABC):
         except UnknownTimeZoneError:
             logger.warning('Warning, server is using an unrecognized timezone %s, will use UTC default', server_tz)
         offsets_differ = datetime.now().astimezone().utcoffset() != datetime.now(tz=self.server_tz).utcoffset()
-        self.apply_server_timezone = apply_server_timezone == 'always' or (
+        self.apply_server_timezone = apply_server_timezone == True or (
                 coerce_bool(apply_server_timezone) and offsets_differ)
         readonly = 'readonly'
         if not self.min_version('19.17'):

### CODE REVIEW

**Comment by genzgd** [comment]
So the underlying problem is that the server has a timezone identified by `Europe/Warsaw`, and your local machine has a timezone that Python identifies as `CEST`.  When the code compares the offsets for now(), those work, and so `clickhouse-connect` uses your local CEST timezone. 

Under the covers, this query:

```
client.query("SELECT toDateTime('2024-02-01')").result_rows
```

returns an epoch timestamp representing a UTC time with no time information, and then converts it to the `CEST` timezone, but February 1st is not in Summer Time, so it's an hour off in that time zone.  What you want is the time in the `CET` timezone, which is not your current local timezone.

Basically, comparing offsets to determine whether to use the local timezone is a bug, because summer time (daylight savings time) dates will always be wrong one way or another.  Accordingly, unless either the client or the server is in UTC, `clickhouse-connect` should never use the local timezone because Python will always use a "summer time aware" timezone like `CET` or `CEST` which will be wrong approximately half the time.

In the next release I'll fix that bug, and use the server timezone in most circumstances.  In the meantime, the quickest workaround is to set `apply_server_timezone` to `always` when calling `get_client`.

KrzysztofPrzygodzki:
Cool.

I don't know if this is important, but I believe that the code is not applying incorrect time zone,
but rather overwrites it entirely. 

For `2024-04-01` (summer time) offset is `7200`, which is correct.

client.query("SELECT timeZoneOffset(toDateTime('2024-04-01'))").result_rows
`[(7200,)]`

client.query("SELECT toDateTime('2024-04-01')").result_rows
`[(datetime.datetime(2024, 4, 1, 0, 0, tzinfo=datetime.timezone(datetime.timedelta(seconds=7200), 'CEST')),)]`

For `2024-02-01` (winter time) offset on the server is `3600`, which is correct. 
However, returned offset is overwritten with summer time.

client.query("SELECT timeZoneOffset(toDateTime('2024-02-01'))").result_rows
`[(3600,)]`

client.query("SELECT toDateTime('2024-02-01')").result_rows
`[(datetime.datetime(2024, 2, 1, 1, 0, tzinfo=datetime.timezone(datetime.timedelta(seconds=7200), 'CEST')),)]`

Thank you for your work and for sharing this tool with community.

PR #354, opened 2024-05-22
Fix precision loss in Float to Decimal conversion

feychenie:
## Summary
Issue raised in [this slack discussion](https://clickhousedb.slack.com/archives/CU478UEQZ/p1716398118092689)

When converting a float value to a decimal field, values like `0.1234567` end up being `0.1234566` in clickhouse.
There seem to be two different ways of fixing this:
- casting the float as a string before passing it to the `Decimal` constructor.
- not casting the float, but rounding the conversion instead of truncating it

This PR uses string casting to fix the issue, as most examples of decimal usages pass values represented as strings to the `Decimal` constructor.

## Checklist

### DIFF
--- a/clickhouse_connect/datatypes/numeric.py
+++ b/clickhouse_connect/datatypes/numeric.py
@@ -291,9 +291,9 @@ class Decimal(ClickHouseType):
             dec = decimal.Decimal
             mult = self._mult
             if self.nullable:
-                write_array(self._array_type, [int(dec(x) * mult) if x else 0 for x in column], dest)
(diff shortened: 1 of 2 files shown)


genzgd:
Thanks for the contribution.  It turns out that Python itself uses the string approach to round floats:  https://github.com/python/cpython/blob/main/Objects/floatobject.c#L980-L984

genzgd:
I'll do a release over the weekend.  There's one other fix I'd like to investigate but I'll release this regardless.

genzgd:
Released.  Thanks again!

Issue #344, opened 2024-04-21
Ability to reduce data exposed via exception


royxact:
Hi team,

We just had a PT done to our system. 
Specifically the part of our system which uses Superset.
The problem is with the error message clickhouse-connect returns.
It includes both the server name and the clickhouse version. This is great in some cases, but no so good for security as end users of superset are not supposed to know these details. Can you add a flag to reduce the error messages to just the minimal SQL error?
![unnamed (1)](https://github.com/ClickHouse/clickhouse-connect/assets/10857260/91d622d1-93a6-4756-a39b-a0880a815a30)

Thanks!

PR #364, opened 2024-06-23
Sercurity issue: remove clickhouse server ip & port in the exception

andy1xx8:
Hi,
The issue is the HTTPDriver return the Clickhouse server sensitive detail: **IP & Port** in the error exception.
Example:
```
Error: HTTPDriver for http://10.15.27.43:8123 returned response code 404)
 Code: 60. DB::Exception: Table test.users_1 does not exist. Maybe you meant test.users? (UNKNOWN_TABLE) (version 23.12.3.40 (official build))
```
Expected:
```
Error: HTTPDriver returned response code 404)
 Code: 60. DB::Exception: Table test.users_1 does not exist. Maybe you meant test.users?. (UNKNOWN_TABLE) (version 23.12.3.40 (official build))
```

This will lead to an issue that the downstream application using this Clickhouse Connect driver, 
They will return the details to the client side and the UI will show the **IP & Port** then the hacker may use this to attack. 
We might run into these 2 cases:
+ They will wrap the exception and not return the detailed error to the client side => This is a bad practice cause the user will not know the message details.
+ They will try to manipulate the message string to search in the string to remove **IP & Port** ==> This is also an ad-hoc fix since the message string may have many different templates and formats, that they can not cover every case.

 ## Summary

## Checklist

### DIFF
--- a/clickhouse_connect/driver/httpclient.py
+++ b/clickhouse_connect/driver/httpclient.py
@@ -426,7 +426,7 @@ class HttpClient(Client):
                         logger.debug('Retrying remotely closed connection')
                         continue
                 logger.warning('Unexpected Http Driver Exception')
-                raise OperationalError(f'Error {ex} executing HTTP request attempt {attempts} {self.url}') from ex
+                raise OperationalError(f'Error {ex} executing HTTP request attempt {attempts}') from ex
             finally:
                 if query_session:
                     self._active_session = None  # Make sure we always clear this
--- a/clickhouse_connect/driver/httpclient.py
+++ b/clickhouse_connect/driver/httpclient.py
@@ -349,7 +349,7 @@ class HttpClient(Client):
         return QuerySummary(self._summary(response))
 
     def _error_handler(self, response: HTTPResponse, retried: bool = False) -> None:
-        err_str = f'HTTPDriver returned response code {response.status})'
+        err_str = f'HTTPDriver for {self.url} returned response code {response.status})'
         try:
             err_content = get_response_data(response)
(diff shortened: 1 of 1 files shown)


genzgd:
Thanks for your submission!  However, instead of always removing that information (which could be useful for multi-client applications), please make that behavior configurable using the new `show_clickhouse_errors` global setting.

andy1xx8:
@genzgd  WWow thanks, if there is a global setting like that I will add a logic to turn off sensitive information  respect the value of show_clickhouse_errors.
How do u think? 

Thanks

genzgd:
That sounds good -- for reference it was added in response to #344 where it came up, but I missed those places where it was exposed directly in the error message.  Here's the v0.7.9 changelog message too: 

```
- The new client keyword argument `show_clickhouse_errors` controls whether the full ClickHouse error (including possibly
sensitive information) is displayed when there is an error in ClickHouse processing.  It defaults to True.  If False,
the simple string 'The ClickHouse server returned an error.' will be displayed.  Closes https://github.com/ClickHouse/clickhouse-connect/issues/344.
```

andy1xx8:
@genzgd 

I have just changed the code logic to respect the config `show_clickhouse_errors`. Please help me to review and merge if it's ok.

We are facing this issues that showing the sensitive info to the client even if we turn off the show_clickhouse_errors

genzgd:
I think you need the same check in the `_error_handler` method where you removed the url before?

andy1xx8:
@genzgd
Yes. Done. Please help me to review and merge if it's ok.

PR #374, opened 2024-07-01
Add contributing guidelines, update docker-compose files, update the CI

slvrtrn:
## Summary

This PR adds the contributing guide with a step-by-step tutorial on how to set up the project.

Additionally, the local ClickHouse instances setup is now more streamlined; in the project root, there is now a Docker Compose file, which contains both pre-configured ClickHouse instances—the "default" plain text auth and the TLS with certificates.

The CI was also updated/simplified a bit. Now, it runs TLS tests with the entire matrix (previously disabled?).

NB: the `clickhouse_ci` needs a proper split; it can be done as a follow-up.


.github/workflows/clickhouse_ci.yml
```diff
@@ -61,18 +52,8 @@ jobs:
...
-          clickhouse/clickhouse-server:head
+        run: CLICKHOUSE_VERSION=head docker compose up -d clickhouse
       - name: Run HEAD tests
         run: pytest tests/integration_tests
-      - name: stop head container
-        run: docker stop clickhouse_head
       - name: remove head container
-        run: docker remove clickhouse_head
-
+        run: docker compose down -v
```
slvrtrn, line 59: IMO, this needs splitting + a proper docker-compose action (like in the `on_push.yml`).

.github/workflows/on_push.yml
```diff
@@ -96,12 +94,19 @@ jobs:
...
       - name: Run tests
         env:
           CLICKHOUSE_CONNECT_TEST_DOCKER: 'False'
           CLICKHOUSE_CONNECT_TEST_FUZZ: 50
           SQLALCHEMY_SILENCE_UBER_WARNING: 1
         run: pytest tests
+      - name: Run TLS tests
+        env:
+          CLICKHOUSE_CONNECT_TEST_TLS: 1
+        run: pytest tests/tls
```
slvrtrn, line 109: This was previously missing. Now, it is possible, as we have a TLS instance in Docker.

docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+    ports:
+      - '10843:8443'
+      - '10840:9440'
+    ulimits:
+      nofile:
+        soft: 262144
+        hard: 262144
+    volumes:
+      - './.docker/clickhouse/single_node_tls/config.xml:/etc/clickhouse-server/config.xml'
+      - './.docker/clickhouse/single_node_tls/users.xml:/etc/clickhouse-server/users.xml'
```
slvrtrn, line 31: The entire setup is mostly taken from CHJS (with the certificates from this project).

tests/integration_tests/conftest.py
```diff
@@ -67,14 +67,14 @@ def test_table_engine_fixture() -> Iterator[str]:
...
     if test_config.docker:
-        run_cmd(['docker-compose', '-f', compose_file, 'down', '-v'])
+        run_cmd(['docker', 'compose', '-f', compose_file, 'down', '-v'])
         sys.stderr.write('Starting docker compose')
-        pull_result = run_cmd(['docker-compose', '-f', compose_file, 'pull'])
+        pull_result = run_cmd(['docker', 'compose', '-f', compose_file, 'pull'])
         if pull_result[0]:
             raise TestException(f'Failed to pull latest docker image(s): {pull_result[2]}')
-        up_result = run_cmd(['docker-compose', '-f', compose_file, 'up', '-d'])
+        up_result = run_cmd(['docker', 'compose', '-f', compose_file, 'up', '-d', 'clickhouse'])
```
slvrtrn, line 77: This is some PyCharm weirdness (does not like `docker-compose`, only works with `docker compose`). Also, could it be a good idea to disable this auto-docker-compose by default? CI does not use it; on a local machine, it's easier to just run `docker compose up -d` beforehand, and that's it. Cause otherwise, without proper healthcheck, it's at least +5 sec of waiting for a single test run (from PyCharm, for example), plus an unnecessary `docker compose down -v` each time.

tests/tls/test_tls.py
```diff
@@ -1,62 +1,39 @@
...
 host = 'server1.clickhouse.test'
 
-"""
-Sample openSSL section for clickhouse server configuration using the certificates in this directory
-
-<openSSL>
-    <server>
-        <certificateFile>clickhouse_test_server.crt</certificateFile>
-        <privateKeyFile>clickhouse_test_server.key</privateKeyFile>
-        <verificationMode>strict</verificationMode>
```
slvrtrn, line 21: Even with the certificates provided in this folder, I could not get it working in strict mode. So, in the `.docker/clickhouse/single_node_tls/config.xml`, it is `relaxed` now.

tests/unit_tests/test_driver/binary.py
```diff
@@ -7,4 +7,4 @@
 6504 6669 7665 036f 6e65 0374 776f 036f  
 6e65 0374 776f 0500 0000 4d00 0000 0500  
 0000 3700 0000 0500 0000 3700 0000  
-"""
\ No newline at end of file
+"""
```
slvrtrn, line 10: PyLint was not happy.

.docker/clickhouse/single_node/config.xml
```diff
@@ -0,0 +1,60 @@
...
+        <partition_by>toYYYYMM(event_date)</partition_by>
+        <flush_interval_milliseconds>1000</flush_interval_milliseconds>
+    </query_log>
+
+    <session_log>
+        <database>system</database>
+        <table>session_log</table>
+    </session_log>
+
+    <http_options_response>
```
genzgd, line 40: When do we see OPTIONS requests?

slvrtrn: Ah, that's a copy-paste from this one: https://github.com/ClickHouse/clickhouse-js/blob/main/.docker/clickhouse/single_node/config.xml#L35-L52

Can remove if this is unnecessary.

tests/integration_tests/conftest.py
genzgd: I have no problem with disabling the automatic docker compose start.  (I generally run tests on a local instance  using different ports anyway, because then I don't have to worry about conflicts ClickHouse docker instances started by my projects.)

genzgd (approved):

PR #375, opened 2024-07-04
AsyncClient wrapper

slvrtrn:
## Summary

This PR adds the `AsyncClient` wrapper over the standard `Client`; it works similarly to how it is shown in [the run_async example](https://github.com/ClickHouse/clickhouse-connect/blob/main/examples/run_async.py); however, the `AsyncClient` uses the `run_in_executor` to run the "blocking" code in a separate thread instead of `asyncio.to_thread` which is 3.9+ only.

TODO: 
- [x] Add a proper per-client `autogenerate_session_id` override
- [x] Update the README and the async example

## Checklist
- [x] For significant changes, documentation in https://github.com/ClickHouse/clickhouse-docs was updated with further explanations or tutorials


genzgd (approved):

slvrtrn:
@genzgd, before merging this:

> Double-check if it's thread-safe

Can you please confirm that we will not run into the thread-safety issues given that `loop.run_in_executor` effectively executes the code in just _some_ thread, which is not necessarily the same, as I understand it? Maybe there are some internal caches or similar things that are worth double-checking?

PR #388, opened 2024-08-21
Fix insert large string

bakwc:
## Summary
Fixed insertion of large string. Currently when inserting large string there is an exception:

```
  File "/app/mysql_ch_replicator/clickhouse_api.py", line 140, in insert
    self.client.insert(table=full_table_name, data=records_to_insert)
  File "/usr/local/lib/python3.12/site-packages/clickhouse_connect/driver/client.py", line 616, in insert
    context.data = data
    ^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/clickhouse_connect/driver/insert.py", line 96, in data
    self.block_row_count = self._calc_block_size()
                           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/clickhouse_connect/driver/insert.py", line 120, in _calc_block_size
    return 1 << (21 - int(log(row_size, 2)))
```


genzgd:
Thanks for the fix!  Just to confirm, this looks like it would break only at strings that are close to 1MB in size, is that about the size where you saw the problem?

genzgd:
Also if you want to bump the version in the `__version__` file I can push this today (or I can do another quick PR for that).

bakwc:
I noticed on 7mb string.

PR #397, opened 2024-09-19
Include the column name in the error message for an unexpected NULL

angusholder:
Previously if you inserted a NULL into a column that isn't Nullable, the error you got was 

> Unable to create Python array. This is usually caused by trying to insert None values into a ClickHouse column that is not Nullable

which is unhelpful for working out *which* column is the problem. I've modified that error path so it can include the column name in the error, like so:

> Failed to write column 'bus_voltage': Unable to create Python array. This is usually caused by trying to insert None values into a ClickHouse column that is not Nullable

I felt like this was a pretty small change so I didn't add tests or file an issue, I hope that's okay.

### DIFF
--- a/clickhouse_connect/driver/common.py
+++ b/clickhouse_connect/driver/common.py
@@ -38,12 +38,13 @@ def array_type(size: int, signed: bool):
     return code if signed else code.upper()
 
 
-def write_array(code: str, column: Sequence, dest: MutableSequence):
+def write_array(code: str, column: Sequence, dest: MutableSequence, ctx):
     """
     Write a column of native Python data matching the array.array code
     :param code: Python array.array code matching the column data type
     :param column: Column of native Python values
     :param dest: Destination byte buffer
+    :param ctx: The InsertContext
     """
     if len(column) and not isinstance(column[0], (int, float)):
         if code in ('f', 'F', 'd', 'D'):
@@ -54,8 +55,8 @@ def write_array(code: str, column: Sequence, dest: MutableSequence):
         buff = struct.Struct(f'<{len(column)}{code}')
         dest += buff.pack(*column)
     except (TypeError, OverflowError, struct.error) as ex:
-        raise DataError('Unable to create Python array.  This is usually caused by trying to insert None ' +
-                        'values into a ClickHouse column that is not Nullable') from ex
+        raise ctx.make_data_error('Unable to create Python array.  This is usually caused by trying to insert None ' +
+                                  'values into a ClickHouse column that is not Nullable') from ex
 
 
 def write_uint64(value: int, dest: MutableSequence):
--- a/clickhouse_connect/driver/common.py
+++ b/clickhouse_connect/driver/common.py
@@ -2,11 +2,13 @@ import array
 import struct
 import sys
 
-from typing import Sequence, MutableSequence, Dict, Optional, Union, Generator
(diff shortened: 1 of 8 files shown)


clickhouse_connect/driver/insert.py
```diff
@@ -198,3 +199,12 @@ def _convert_numpy(self, np_array):
                 data[ix] = data[ix].tolist()
         self.column_oriented = True
         return data
+
+    def start_column(self, name: str):
```
angusholder, line 203: This gets called during the insert to tell us what the current column being inserted is, so I stored its name in case we run into an error for that column

clickhouse_connect/driver/insert.py
```diff
@@ -198,3 +199,12 @@ def _convert_numpy(self, np_array):
...
+        super().start_column(name)
+        self._column_name = name
+
+    def make_data_error(self, error_message: str) -> DataError:
+        if self._column_name is not None:
```
angusholder, line 208: I don't know if it's possible for us to be handling a column insert without start_column() having been called, but I handled `None` just in case

clickhouse_connect/driver/common.py
```diff
@@ -54,8 +55,8 @@ def write_array(code: str, column: Sequence, dest: MutableSequence):
         buff = struct.Struct(f'<{len(column)}{code}')
         dest += buff.pack(*column)
     except (TypeError, OverflowError, struct.error) as ex:
-        raise DataError('Unable to create Python array.  This is usually caused by trying to insert None ' +
-                        'values into a ClickHouse column that is not Nullable') from ex
+        raise ctx.make_data_error('Unable to create Python array.  This is usually caused by trying to insert None ' +
```
angusholder, line 58: This was the only error I _really_ needed improving, but I modified the other places where DataError could be raised too, in string.py

clickhouse_connect/driver/common.py
```diff
@@ -38,12 +38,13 @@ def array_type(size: int, signed: bool):
     return code if signed else code.upper()
 
 
-def write_array(code: str, column: Sequence, dest: MutableSequence):
+def write_array(code: str, column: Sequence, dest: MutableSequence, ctx):
```
genzgd, line 41: I think you can add the ctx Type here too?

clickhouse_connect/driver/insert.py
genzgd: It's not possible, so it's safe to remove the None check.

genzgd (commented):
Thanks so much, I agree it's hard to track down obscure insert errors and this will help.  Stashing the column name in InsertContext makes a lot of sense, btw -- I haven't looked deeply, but maybe we add that to the BaseQueryContext instead since it doesn't hurt to have it around for queries as well.

angusholder:
> Thanks so much, I agree it's hard to track down obscure insert errors and this will help. Stashing the column name in InsertContext makes a lot of sense, btw -- I haven't looked deeply, but maybe we add that to the BaseQueryContext instead since it doesn't hurt to have it around for queries as well.

Sounds like a good idea, I've moved it to BaseQueryContext as you suggest

genzgd:
Thanks again!   I'm hoping to do another release next week that will include this.

angusholder:
No problem! Great to hear, thanks

PR #442, opened 2024-12-20
JWT auth support

slvrtrn:
## Summary

This PR adds support for JWT authentication (ClickHouse Cloud feature).

- [x] Add JWT to the client configuration
- [x] Allow to override the token

## Checklist


slvrtrn:
As JWT expires over time, we need a way to override it in the client.

@genzgd, perhaps a `set_access_token` method on the client instance could be a good idea? An alternative is to add authorization override on all the public methods, which is probably too much and won't be too convenient to use. WDYT?

genzgd:
`set_access_token` makes sense.  If we wanted to get fancy we could make the access token constructor parameter take either a string or a function/hook that returns a string so the user could use their own implementation for refreshing?

mzitnik (approved):
LGTM

slvrtrn:
I added the `set_access_token` method for starters. We can provide more exquisite support with internal timers as a follow-up.

genzgd (approved):

PR #445, opened 2024-12-31
Support set ciphers

caicancai:
## Summary

https://quentin.pradet.me/blog/fixing-sslv3_alert_handshake_failure-with-urllib3-20.html
urllib3 supports set_ciphers

## Checklist

### DIFF
--- a/clickhouse_connect/driver/httpclient.py
+++ b/clickhouse_connect/driver/httpclient.py
@@ -73,7 +73,8 @@ class HttpClient(Client):
                  apply_server_timezone: Optional[Union[str, bool]] = None,
                  show_clickhouse_errors: Optional[bool] = None,
                  autogenerate_session_id: Optional[bool] = None,
-                 tls_mode: Optional[str] = None):
+                 tls_mode: Optional[str] = None,
+                 cipher: Optional[str] = None):
         """
         Create an HTTP ClickHouse Connect client
         See clickhouse_connect.get_client for parameters
@@ -105,6 +106,8 @@ class HttpClient(Client):
                     if options['verify']:
                         options['assert_hostname'] = server_host_name
                     options['server_hostname'] = server_host_name
+                if cipher:
+                    options['cipher'] = cipher
                 self.http = get_pool_manager(https_proxy=https_proxy, **options)
                 self._owns_pool_manager = True
         if not self.http:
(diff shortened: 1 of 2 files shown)


caicancai:
If you need me to update the documentation and tests, please let me know. I am not very clear about the rules of the ClickHouse community.

genzgd:
Thanks for the contribution!  I hesitate to add yet another parameter to client instantiation, especially something that very few people will use (I think most users are fine with the standard cipher suite).  It seems like this is an issue that might be better handled by just creating a custom pool_manager instead?

I'll give it some thought.  In the meantime we can't accept code without a signed CLA.  Also, an integration test to make sure that creating the SSL context with a specific cipher would be helpful.

caicancai:
> Thanks for the contribution! I hesitate to add yet another parameter to client instantiation, especially something that very few people will use (I think most users are fine with the standard cipher suite). It seems like this is an issue that might be better handled by just creating a custom pool_manager instead?
> 
> I'll give it some thought. In the meantime we can't accept code without a signed CLA. Also, an integration test to make sure that creating the SSL context with a specific cipher would be helpful.

@genzgd Thank you for your reply. This parameter is useful for me. When the length of the clickhosue certificate is not 2048 bytes but 1024 bytes, clickhouse-connect will not work, causing me to manually set_ciphers

If you have a better solution, please tell me and I can study it

If you accept my solution, I am happy to add integration tests and documentation

genzgd:
I understand the issue.  You can solve the problem without changing existing code by creating your own pool manager:

```python
from urllib3.util import create_urllib3_context

from clickhouse_connect.driver.httputil import get_pool_manager

cipher = 'DHE-RSA-AES128-SHA'
ctx = create_urllib3_context()
ctx.load_default_certs()
ctx.set_ciphers(cipher)

pool_mgr = get_pool_manager(ssl_context=ctx)

client = get_client(pool_mgr=pool_mgr)
```

Is that good enough for your use case?

caicancai:
> I understand the issue. You can solve the problem without changing existing code by creating your own pool manager:
> 
> ```python
> from urllib3.util import create_urllib3_context
> 
> from clickhouse_connect import get_client
> from clickhouse_connect.driver.httputil import get_pool_manager
> 
> cipher = 'DHE-RSA-AES128-SHA'
> ctx = create_urllib3_context()
> ctx.load_default_certs()
> ctx.set_ciphers(cipher)
> 
> pool_mgr = get_pool_manager(ssl_context=ctx)
> 
> client = get_client(pool_mgr=pool_mgr)
> ```
> 
> Is that good enough for your use case?

thanks, let me try it

caicancai:
It doesn't look like it's possible, I'll take a look at the code over the weekend to see if there's a better way

caicancai:
Temporarily close this PR and look for a better solution

PR #448, opened 2025-01-07
fix: add default value for access token in HttpClient init

lukasthalerINNIO:
## Summary
0.8.12 introduced a new `access_token` parameter to the `clickhouse_connect.driver.HttpClient` class that is marked as `Optional[str]` but has no default value. It is properly defaulted in `create_client`, but if you directly create an instance of `HttpClient` in 0.8.12, it fails due to the missing new argument

This PR adds a `None` default value to the `HttpClient` `__init__` to mirror `create_client` and fix the issue

## Checklist
- [x] A human-readable description of the changes was provided to include in CHANGELOG: Set default value for the optional `access_token` parameter of the `HttpClient.__init__` method

### DIFF
--- a/clickhouse_connect/driver/httpclient.py
+++ b/clickhouse_connect/driver/httpclient.py
@@ -54,7 +54,7 @@ class HttpClient(Client):
                  username: str,
                  password: str,
                  database: str,
-                 access_token: Optional[str],
+                 access_token: Optional[str] = None,
                  compress: Union[bool, str] = True,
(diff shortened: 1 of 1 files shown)


lukasthalerINNIO:
A quick release would be much appreciated. We use uncapped `clickhouse-connect` dependencies a fair bit and all new builds are crashing due to this. I've put out a PSA to the team, but a new release would make our lives a lot easier

PR #450, opened 2025-01-09
Make all None by default arguments accept None value

orian:
## Summary

Some arguments take str type hint, but default value is None.


genzgd:
Thanks for the PR!

Issue #424, opened 2024-11-17
Memory leaks AsyncClient


ZwickVitaly:
### Describe the bug
I need to write a massive (2b rows per 24 hours) amount of data every day. 
Application does 1m requests per city. (4 cities)
Evey request gets 500 rows of info.
Every 4 requests I write rows to clickhouse db using clickhouse-connect -> 2k products per write
Got 8 celery processes, each does 125k requests part of all
Each process does 4 requests async.gather -> gets 4 lists of 500 rows, combines them to 2k list
There goes client.insert

Every 10 minutes SWAP memory increases by 50mb, BUT **Database swap usage stays on same level always (70MB)**
smem shows, that celery processes grow in SWAP memory.

I was worried that aiohttp leaks, but **removing insert and doing just requests does not make memory grow at all**

**so the problem is insert**

### Steps to reproduce
1. Run process, that writes something with AsyncClient.insert in a loop 
2. Wait 
3. See swap memory growth

### Expected behaviour
No swap memory growth

Seems to me, that client does not clear QueryResult or QueryResultSummary properly, but I am not smart enough to figure it out.

### Code example
made custom async context manager for getting connect
```python
@asynccontextmanager
async def get_async_connection() -> AsyncClient:
    session = AsyncSession(CLICKHOUSE_CONFING)
    async with session as client:
        yield client


class AsyncSession:

    def __init__(self, clickhouse_config):
        self.config = clickhouse_config

    async def __aenter__(self) -> AsyncClient:
        self.client = await clickhouse_connect.get_async_client(**self.config)
        return self.client

    async def __aexit__(self, exc_type, exc_val, exc_tb):
        self.client.close()
```
this is request function
```python
async def get_city_result(city, date, requests, request_batch_no, client=None):
    requests_list = [r for r in requests if not r[1].isdigit() or "javascript" not in r[1]]
    batch_size = 4
    prev = 0
    async with ClientSession() as http_session:
        async with get_async_connection() as client:
            for batch_i in range(batch_size, len(requests_list) + 1, batch_size):
                request_batch = requests_list[prev:batch_i]
                requests_tasks = [
                    asyncio.create_task(
                        get_r_data(
                            r=r,
                            city=city,
                            date=date,
                            http_session=http_session,
                        )
                    )
                    for r in request_batch
                ]
                product_batches: tuple = await asyncio.gather(*requests_tasks)
                prev = batch_i
                full_res = []
                for batch in product_batches:
[...]

PR #457, opened 2025-01-21
Fix memory leak in AsyncClient

pufit:
## Summary
It's known that ThreadPoolExecutor doesn't deallocate memory without .shutdown being called. Python's GC doesn't stop running thread pools, so the code like
```python
while True:
    client = await get_async_client(...)
    ...
    client.close()
```
will be leaking memory since each new AsyncClient instance will have a new executor.

Closes https://github.com/ClickHouse/clickhouse-connect/issues/424

## Checklist

### DIFF
--- a/clickhouse_connect/driver/asyncclient.py
+++ b/clickhouse_connect/driver/asyncclient.py
@@ -69,6 +69,7 @@ class AsyncClient:
         Subclass implementation to close the connection to the server/deallocate the client
         """
         self.client.close()
+        self.executor.shutdown(wait=False)
 
     async def query(self,
                     query: Optional[str] = None,
--- a/clickhouse_connect/driver/asyncclient.py
+++ b/clickhouse_connect/driver/asyncclient.py
@@ -64,12 +64,12 @@ class AsyncClient:
         """
         return self.client.min_version(version_str)
 
-    def close(self):
+    async def close(self):
         """
         Subclass implementation to close the connection to the server/deallocate the client
         """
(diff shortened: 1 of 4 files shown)


clickhouse_connect/driver/asyncclient.py
```diff
@@ -64,11 +64,12 @@ def min_version(self, version_str: str) -> bool:
         """
         return self.client.min_version(version_str)
 
-    def close(self):
+    async def close(self):
```
genzgd, line 67: Does this have to be async?  Can we just call the executor.shutdown method synchronously?

genzgd (commented):
Thanks for digging into this!  My only question is whether this can be simpler by just shutting down the executor synchronously.  It looks like you added the `__aenter__` and `__aexit__` methods to ensure this autocloses when using it as an AsyncContext?

pufit:
> Thanks for digging into this! My only question is whether this can be simpler by just shutting down the executor synchronously. It looks like you added the `__aenter__` and `__aexit__` methods to ensure this autocloses when using it as an AsyncContext?

This way is more Pythonic. Usually, you open and close clients inside the event loop. If you try to wait for executor shutdown in sync mode it will hang the loop. If you close the thread pool without wait=True, you won't be able to guarantee the graceful shutdown of your application.

Here is an example of how it is usually done in other libraries: https://github.com/aio-libs/aiohttp/blob/master/examples/curl.py#L11

PR #506, opened 2025-06-18
Fix some issues with cursor behavior

### LINKED ISSUE #505: incorrect pagination in Cursor.fetchall and fetchmany
### Describe the bug

The bugs are originally described in: https://github.com/ClickHouse/clickhouse-connect/pull/503.

They are:

- `fetchall` returned rows that had already been consumed (pep-249 specifies all remaining)
- `fetchmany` returned all remaining rows, regardless of the `size` parameter

In user's case, these bugs surfaced by causing duplication of all rows in query results when using a clickhouse connection in [jupysql](https://pypi.org/project/jupysql/), if that context is interesting.

### Expected behaviour

- `fetchall` should return just _remaining_ rows
- `fetchmany` should obey `size` parameter

joe-clickhouse:
## Summary
This PR fixes some issues with cursor behavior (thanks @celeste-sinead). It updates behavior to be more consistent with [PEP-249](https://peps.python.org/pep-0249/), specifically around [cursor methods](https://peps.python.org/pep-0249/#cursor-methods):
- **Reset cursor location after performing an execute** (previously cursor position was not reset)
- **Fix behavior of `fetchall` to only return rows from the current cursor location** (previously returned all rows regardless of cursor position)
- Fixes logic of `fetchmany` to respect size parameter.
- Adds file specifically for testing cursor behavior.

Note: One aspect still inconsistent with PEP-249 is the default size parameter in the `fetchmany` implementation. Currently it defaults to `-1` which fetches all remaining rows. According to the PEP, it should fall back fall back to the `self.arraysize` value, which defaults to `1`. This would be a breaking change. Since we're still under v1.0 it would be ok to change. To address in the future, we can open another issue.

Closes #505

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # ClickHouse Connect ChangeLog
 
+## UNRELEASED
+
+## Improvements
+- Added a standalone test file (`tests/unit_tests/test_driver/test_cursor.py`) for testing cursor behavior
+
+## Bug Fixes
+- Reset cursor location after performing an execute.
+- Fix behavior of `fetchall` to only return rows from the current cursor location.
+- Fixes logic of `fetchmany` to respect size parameter.
+
 ### WARNING -- Breaking change for AsyncClient close()
 The AsyncClient close() method is now async and should be called as an async function.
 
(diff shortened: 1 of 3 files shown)


genzgd (approved):
LGTM.  If you think there really is a breaking change it should go at the top of the changelog for a few patch releases.

PR #511, opened 2025-06-27
Joe/508 please propagate error messages from http 400 at least

### LINKED ISSUE #508: Please propagate error messages from HTTP 400 (at least)
### Steps to reproduce
1. I had clickhouse running in a kubernetes cluster and port-forwarded to it
2. I verified that `clickhouse client --no-secure ...` could connect just fine to the port-forward
3. Then I tried to connect from python, `clickhouse_connect.get_client(...)` etc.
4. Actual behavior is an error inside the clickhouse_connect core, complaining that it can't `SELECT version(), timezone()`:

```
  File "/Users/cdrost/repos/xyz/abc/clickhouse_connect/driver/client.py", line 75, in _init_common_settings
    tuple(self.command('SELECT version(), timezone()', use_database=False))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/cdrost/repos/xyz/abc/clickhouse_connect/driver/httpclient.py", line 368, in command
    response = self._raw_request(payload, params, headers, method, fields=fields, server_wait=False)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/cdrost/repos/xyz/abc/clickhouse_connect/driver/httpclient.py", line 483, in _raw_request
    self._error_handler(response)
  File "/Users/cdrost/repos/xyz/abc/clickhouse_connect/driver/httpclient.py", line 402, in _error_handler
    raise OperationalError(err_str) if retried else DatabaseError(err_str) from None
clickhouse_connect.driver.exceptions.DatabaseError: HTTPDriver for http://localhost:21212 returned response code 400
```

### Expected behaviour

When vendoring the code into my app and adding printf debugging I find the error very easily:

```
--> _raw_request <clickhouse_connect.driver.httpclient.HttpClient object at 0x1009aa310> SELECT version(), timezone() {} {} POST 0 False False None None
--> response 400 b'Port 9000 is for clickhouse-client program\r\nYou must use port 8123 for HTTP.\r\n' HTTPHeaderDict({}) /?
```

I am kubectl port-forwarding to the wrong port. However **the HTTP response text here should not be eaten** by clickhouse_connect, it made a 5 minute error take like hours of trying different things (in particular the one result for this on Stack Overflow is an HTTP/HTTPS confusion error so there was a lot of messing with that)

joe-clickhouse:
## Summary
Improved error reporting and includes minor improvements to the httpclient's error handler. It now:

1. Always attempts body retrieval.
2. Always includes server's response when available.
3. Ensures response object is always closed.
4. Has added unit and integration tests.

Closes #508

## Checklist

### DIFF
--- a/clickhouse_connect/driver/httpclient.py
+++ b/clickhouse_connect/driver/httpclient.py
@@ -380,25 +380,47 @@ class HttpClient(Client):
         return QuerySummary(self._summary(response))
 
     def _error_handler(self, response: HTTPResponse, retried: bool = False) -> None:
-        if self.show_clickhouse_errors:
+        """
+        Handles HTTP errors. Tries to be robust and provide maximum context.
+        """
+        try:
+            body = ""
+            # Always try to read the response body for context.
             try:
-                err_content = get_response_data(response)
+                # get_response_data reads body and decodes it for the error message
+                raw_body = get_response_data(response)
+                body = common.format_error(
+                    raw_body.decode(errors="backslashreplace")
+                ).strip()
             except Exception:  # pylint: disable=broad-except
-                err_content = None
-            finally:
-                response.close()
+                # If we can't read or decode the body, we'll proceed without it
+                logger.warning("Failed to read error response body", exc_info=True)
 
-            err_str = f'HTTPDriver for {self.url} returned response code {response.status}'
-            err_code = response.headers.get(ex_header)
-            if err_code:
-                err_str = f'HTTPDriver for {self.url} received ClickHouse error code {err_code}'
-            if err_content:
-                err_msg = common.format_error(err_content.decode(errors='backslashreplace'))
-                if err_msg.startswith('Code'):
-                    err_str = f'{err_str}\n {err_msg}'
-        else:
-            err_str = 'The ClickHouse server returned an error.'
+            # Build the error message
+            if self.show_clickhouse_errors:
+                err_code = response.headers.get(ex_header)
+                if err_code:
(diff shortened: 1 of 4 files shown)


clickhouse_connect/driver/httpclient.py
```diff
@@ -380,25 +380,47 @@ def command(self,
...
+                err_code = response.headers.get(ex_header)
+                if err_code:
+                    # Prioritize the specific ClickHouse exception code if it exists.
+                    err_str = f"Received ClickHouse exception, code: {err_code}"
+                else:
+                    # Otherwise, just use the generic HTTP status
+                    err_str = f"HTTP driver received HTTP status {response.status}"
+
+                if body:
+                    # Always append the body if it exists
```
genzgd, line 410: I think we do the err_content.decode for a reason here -- in particular I remember something about UTF-8 values in the body that don't get properly shown (or error) if we don't do the decode.  There probably should have a been a test for it, but I suspect I got lazy.

genzgd: Yes, there's a specific PR for it:  https://github.com/ClickHouse/clickhouse-connect/pull/19

genzgd (approved):

PR #522, opened 2025-07-21
Joe/510 support bfloat16

joe-clickhouse:
## Summary
Adds support for BFloat16 types.

## Checklist


joe-clickhouse:
Yeah, no native 16-bit support unfortunately. And yeah, in the reconstruction direction when we go from the [BFloat16](https://en.wikipedia.org/wiki/Bfloat16_floating-point_format) -> Python 64-bit float the lower bits are always just zero padded. We're always going to lose that precision when we truncate the mantissa in other direction but that's sorta the design intent and purpose of the BFloat16 in general. The idea is that you can drastically cut your I/O and memory use by storing a fraction of the amount of data by giving up a little precision as an acceptable tradeoff for ML applications). High-precision numerical calculations are not the intended use case here. And it looks like the [request](https://github.com/numpy/numpy/issues/17809) for numpy to support BFloat16 has been around for years and has essentially been closed as a won't do.

PR #541, opened 2025-08-20
support for sqalchemy 2.x

### LINKED ISSUE #263: Support SQLAlchemy 2.x
We should probably provide limited support for SQLAlchemy 2.x, at least for queries/datatypes.

joe-clickhouse:
## Summary
- Provides support for sqlalchemy 2. Supported version range is now sqlalchemy>=1.4.40,<3.0.
- Removes support for sqlalchemy 1.3 as it [has reached](https://www.sqlalchemy.org/download.html#status-desc-eol) EOL.
- Updates test requirements to pin sqa version to >=1.4.40,<2.0 and adds a new separate CI step to also run all sqalchemy tests against 2.x

Closes #263

## Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -64,7 +64,7 @@ def run_setup(try_c: bool = True):
             'lz4'
         ],
         extras_require={
-            'sqlalchemy': ['sqlalchemy>1.3.21,<2.0'],
+            'sqlalchemy': ['sqlalchemy>=1.4.40,<2.0'],
             'numpy': ['numpy'],
             'pandas': ['pandas'],
             'arrow': ['pyarrow'],
@@ -80,7 +80,6 @@ def run_setup(try_c: bool = True):
             'Development Status :: 4 - Beta',
             'Intended Audience :: Developers',
             'License :: OSI Approved :: Apache Software License',
-            'Programming Language :: Python :: 3.8',
             'Programming Language :: Python :: 3.9',
             'Programming Language :: Python :: 3.10',
             'Programming Language :: Python :: 3.11',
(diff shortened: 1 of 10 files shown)


setup.py
```diff
@@ -64,7 +64,7 @@ def run_setup(try_c: bool = True):
             'lz4'
         ],
         extras_require={
-            'sqlalchemy': ['sqlalchemy>1.3.21,<2.0'],
+            'sqlalchemy': ['sqlalchemy>=1.4.40,<2.0'],
```
genzgd, line 67: This doesn't include SQLAlchemy 2.0?

tests/test_requirements.txt
```diff
@@ -2,7 +2,7 @@ pytz
 urllib3>=1.26
 setuptools
 certifi
-sqlalchemy>1.3.21,<2.0
+sqlalchemy>=1.4.40,<2.0
```
genzgd, line 5: Should we test both 1.4.x and 2.0?

genzgd (approved):

setup.py
joe-clickhouse: Good catch! Thanks. Should be `sqlalchemy>=1.4.40,<3.0`. I'll fix this.

tests/test_requirements.txt
joe-clickhouse: Yes, I added a new step in on-push.yml to test 2.x--similar to the pandas 1.x compatibility testing. But thinking about it now, I think I should run the bulk test matrix with the more modern 2.x and then add a single separate test for 1.x. Will fix.

genzgd: Ah, I missed that, but yeah, matrix testing against 2.x sounds better.

Issue #342, opened 2024-04-18
Send server-side query parameters in POST body instead of in URL


aparshin:
Clickhouse connect supports server-side parameters of query requests, but always sends them as part of URI (particularly, as query parameters). If the server-side parameters are big enough, URL becomes too long and might be rejected by intermediate servers.

Clickhouse server starting from some version supports sending parameters for server-side binding as part of POST request's body: https://github.com/ClickHouse/ClickHouse/issues/8842

It would be great to always send server-side parameters as part of body instead of query parameters in URL.

genzgd:
I had no idea this was possible (since it's not mentioned in the ClickHouse docs.). I think it's probably reasonable, but there are some tradeoffs -- sending everything as a Form to support this is kind of ugly, it adds complexity, and I'm pretty sure it's incompatible with the External Data approach.   (Also, sending huge urls with very long parameter values probably means you're doing something really inefficient in your queries.)

aparshin:
@genzgd In our case, we just query rows with the exact match of a string column to a given string, but the problem is that we can't control size of that string. Sometimes it happens to be huge, which gives us an error from one of intermediate servers.

Generally speaking, putting something with uncontrolled length into URL is a bad practice. HTTP POST with its unlimited body size is a commonly used workaround to prevent such situations. From this point of view, moving all the parameters into body looks like a good improvement to me. Of course, I don't know how it fits to internal architecture. Also, not sure starting from what version this feature is supported at the server...

Meanwhile, the workaround we are going to implement is to use client-side parameters binding. In this case, all big strings will go to query POST body. But ideally we'd like to use sever-side binding.

genzgd:
I agree with you philosophically about bloating the URL -- some time back we moved the query itself into the POST body for exactly that reason.  I'm just not thrilled about using Form fields to do it with the current ClickHouse implementation, and it does add complexity, so this may not happen soon without an outside PR.

Side note -- having query parameters of uncontrolled size is also bad practice. :)

aparshin:
@genzgd Thank you for your feedback, anyway. Not sure that I'll be able to propose a PR for this....

> Side note -- having query parameters of uncontrolled size is also bad practice. :)

I agree that it poses different problems, like the one we are discussing, but I just don't see any alternatives at client-side...

PR #543, opened 2025-08-23
feat(httpclient): add ability to form encode parameters

sanjams2:
== Motivation ==

Support form encoding of query params

== Details ==

This addresses https://github.com/ClickHouse/clickhouse-connect/issues/342

Since https://github.com/ClickHouse/ClickHouse/issues/8842, query parameters can now be sent to ClickHouse when using HTTP via the body as form encoded paramters. This change adds the ability to enable this feature at a client level. Making this available per query (i.e. via the method parameters of the query methods) could be added as a future improvement, but this would require a larger refactor. So for now we are making it a client wide setting AND only available for querying (and not inserting).

== Testing ==

Added and ran integration tests

== Checklist ==

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ The supported method of passing ClickHouse server settings is to prefix such arg
 
 ## UNRELEASED
 - Added support for form encoding query parameters when using HTTP interface. This addresses [#342](https://github.com/ClickHouse/clickhouse-connect/issues/342). Query parameters can now be sent as form-encoded data in the request body by setting `form_encode_query_params=True` when creating the client. This is particularly useful for queries with large parameter payloads that might exceed URL length limits.
+- Added Polars support for Arrow-based query and insert methods (query_df_arrow, query_df_arrow_stream, insert_df_arrow). This initial implementation provides basic dataframe conversion through the Arrow format, similar to how we support the pyarrow-backed pandas dataframes. Closes [#111](https://github.com/ClickHouse/clickhouse-connect/issues/111) and [#542](https://github.com/ClickHouse/clickhouse-connect/issues/542)
 - Added support for SQLAlchemy 2.x. The minimum required version is 1.4.40. Closes [#263](https://github.com/ClickHouse/clickhouse-connect/issues/263)
   - **WARNING: BREAKING CHANGE**: Removed support for sqlalchemy 1.3 which reached its EOL in 2021.
 - Added support for querying/inserting pyarrow-backed DataFrames:
@@ -46,6 +47,7 @@ The supported method of passing ClickHouse server settings is to prefix such arg
 - Fixed an AttributeError on `http.client` when importing `clickhouse_connect` under certain circumstances
 - Added support for Nullable(JSON) types
 - Added support for BFloat16 types
+-- Replace the use of deprecated datetime.utcfromtimestamp
 
 ## 0.8.18, 2025-06-24
 
--- a/clickhouse_connect/datatypes/temporal.py
+++ b/clickhouse_connect/datatypes/temporal.py
@@ -2,12 +2,14 @@ import pytz
 
 import array
 from datetime import date, datetime, tzinfo, timedelta, time
+
 from typing import Union, Sequence, MutableSequence, Any, NamedTuple, Optional
 from abc import abstractmethod
 import re
 
 from clickhouse_connect.datatypes.base import TypeDef, ClickHouseType
 from clickhouse_connect.common import get_setting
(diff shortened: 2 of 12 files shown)


clickhouse_connect/driver/httpclient.py
```diff
@@ -541,15 +554,27 @@ def _prep_raw_query(self, query: str,
         params = self._validate_settings(settings or {})
         if use_database and self.database:
             params['database'] = self.database
-        params.update(bind_params)
-        if external_data:
+        fields = {}
+        # Setup query body
+        if self.form_encode_query_params or external_data:  # form encoded body
             if isinstance(final_query, bytes):
                 raise ProgrammingError('Cannot combine binary query data with `External Data`')
```
joe-clickhouse, line 561: Is this logic correct here? As it stands, the condition will pass even if `external_data` is absent, but the message will mention it. Also, I think we should only disallow bytes when we're forced to put the `query` in the URL, that is when `external_data` _is_ present _and_ we're not form-encoding as that's the only time bytes would end up in `params["query"]`. Correct logic should be something like:

```python
if external_data and not self.form_encode_query_params and isinstance(final_query, bytes):
    raise ProgrammingError("Binary query cannot be placed in URL when using External Data; enable form encoding.")

```

clickhouse_connect/driver/httpclient.py
```diff
@@ -211,9 +213,9 @@ def _query_with_context(self, context: QueryContext) -> QueryResult:
         if self.protocol_version:
             params['client_protocol_version'] = self.protocol_version
             context.block_info = True
-        params.update(context.bind_params)
         params.update(self._validate_settings(context.settings))
         if not context.is_insert and columns_only_re.search(context.uncommented_query):
+            params.update(context.bind_params)
```
joe-clickhouse, line 218: It looks like even if `form_encode_query_params=True`, this right here will just push the bound params into the URL. So for schema-probe queries (e.g. if someone just tacks a `LIMIT 0` to the end of a complex existing query that has a large param payload) this branch could potentially fail while the non-`LIMIT 0` path would not. Plus, it's inconsistent to silently bypass it here. So to fix this here, I think we should mirror the "normal query" behavior implemented a little further down from here.

clickhouse_connect/driver/httpclient.py
sanjams2: Ah yeah error message should be updated. But your proposal makes sense to me. Ill update the logic.

clickhouse_connect/driver/httpclient.py
sanjams2: Makes sense. Will update

clickhouse_connect/driver/httpclient.py
```diff
@@ -211,11 +214,17 @@ def _query_with_context(self, context: QueryContext) -> QueryResult:
...
-                                         params, headers, retries=self.query_retries)
+            # Mirror normal query behavior for form encoding
+            if self.form_encode_query_params:
+                fields = {'query': f'{context.final_query}\n FORMAT JSON'}
+                fields.update(context.bind_params)
+                response = self._raw_request(bytes(), params, headers, retries=self.query_retries, fields=fields)
+            else:
+                params.update(context.bind_params)
+                response = self._raw_request(f'{context.final_query}\n FORMAT JSON',
+                                             params, headers, retries=self.query_retries)
```
joe-clickhouse, line 227: To make this `LIMIT 0` path mirror the non-`LIMIT 0` path below, I think we're still missing the case of external data here. Granted, I think this is a very rare edge case, but for consistency I think it's worth including.

```suggestion
            # Mirror normal query behavior for form encoding
            fmt_json_query = f'{context.final_query}\n FORMAT JSON'
            if self.form_encode_query_params:
                fields = {'query': fmt_json_query}
                fields.update(context.bind_params)
                if context.external_data: # Deal with form encoding + external data
                    params.update(context.external_data.query_params)
                    fields.update(context.external_data.form_data)
                response = self._raw_request(bytes(), params, headers, retries=self.query_retries, fields=fields)
            else:
                if context.external_data: # Deal with external data even without form encoding
                    fields = context.external_data.form_data
                    params.update(context.bind_params)
                    params.update(context.external_data.query_params)
                    params['query'] = fmt_json_query
                    response = self._raw_request(bytes(), params, headers, retries=self.query_retries, fields=fields)
                else: # Legacy behavior (plain body, bind params in URL)
                    params.update(context.bind_params)
                    response = self._raw_request(f'{context.final_query}\n FORMAT JSON',
                                             params, headers, retries=self.query_retries)

joe-clickhouse (approved):
LGTM

PR #547, opened 2025-09-02
Joe/546 add library to user agent

### LINKED ISSUE #546: Add library to User-Agent
To get better insight into where to spend development time and effort supporting libraries like Pandas, Polars, and SQLAlchemy, we need to understand which versions our users have adopted.

We will add a simple mechanism to dynamically prepend the product name and version (e.g., pandas/2.2.1) to the User-Agent string when a relevant API method is called. This option will be default on, with a common setting to opt out.

The User-Agent will change from this:
`clickhouse-connect/0.8.18 (lv:py/3.11.5; os:darwin)`

To this, after a `query_df()` call, for example:
`pandas/2.2.1 clickhouse-connect/0.8.18 (lv:py/3.11.5; os:darwin)`

joe-clickhouse:
## Summary
Adds 3rd party integration tags to the user agent. Prior to this the User-Agent would look like:

```clickhouse-connect/0.8.18 (lv:py/3.12.11; mode:sync; os:darwin)```

If the common setting `send_integration_tags` is set to `True` and the user uses the sqlalchemy integration capabilities, for example, the User-Agent would be updated to:

```clickhouse-connect/0.8.18 (sqlalchemy/2.0.43; lv:py/3.12.11; mode:sync; os:darwin)```

This will greatly help us to determine what kind of development effort should be spent supporting 3rd party integration libraries like pandas, polars, and SQLAlchemy and what versions of those libraries users have adopted.

Closes #546

## Checklist

### DIFF
--- a/clickhouse_connect/dbapi/connection.py
+++ b/clickhouse_connect/dbapi/connection.py
@@ -29,6 +29,10 @@ class Connection:
                                     secure=secure,
                                     dsn=dsn,
                                     generic_args=kwargs)
+        try:
+            self.client._add_integration_tag("sqlalchemy")
+        except Exception: # pylint: disable=broad-exception-caught
+            pass
         self.timezone = self.client.server_tz
 
     def close(self):
--- a/clickhouse_connect/driver/httpclient.py
+++ b/clickhouse_connect/driver/httpclient.py
@@ -3,6 +3,8 @@ import json
 import logging
 import re
 import uuid
+from importlib import import_module
+from importlib.metadata import version as dist_version
 from base64 import b64encode
 from typing import Optional, Dict, Any, Sequence, Union, List, Callable, Generator, BinaryIO
 from urllib.parse import urlencode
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: clickhouse_connect/dbapi/connection.py [hunk 1]
```diff
@@ -29,6 +29,10 @@ def __init__(self,
                                     secure=secure,
                                     dsn=dsn,
                                     generic_args=kwargs)
+        try:
```
genzgd, line 32: Doesn't seem necessary to catch an exception here.  We have full control over the _add_integration_tag method and it shouldn't throw anything.

clickhouse_connect/driver/httpclient.py
```diff
@@ -595,6 +598,51 @@ def _prep_raw_query(self, query: str,
             fields = None
         return body, params, fields
 
+    # pylint: disable=broad-exception-caught
+    def _add_integration_tag(self, name: str):
+        """
```
genzgd, line 603: This seems like a lot of logic to go through on every call.   It's not a big deal, but if we're going to add dynamic content, it seems like the client should store the pieces needed to build the user agent instead of the full raw User-Agent string, and just put those pieces together on each call instead of retrieving and resetting the User Agent header.  Those pieces could also be stored in the Client superclass instead, so we don't have to redo this for Native protocol (someday).

genzgd (approved):

clickhouse_connect/driver/httpclient.py
joe-clickhouse: On second thought here, I think I might just leave as it for now and revisit later as an enhancement. In its current form, it only goes through the complex logic the first time a new library is added and then short-circuits thereafter.

PR #548, opened 2025-09-05
add rename_response_column option

### LINKED ISSUE #228: Support new setting `rename_response_column`
In **clickhouse-java**, there is a setting `rename_response_column` for renaming the column's names https://github.com/ClickHouse/clickhouse-java/pull/923/files
But the config is not support yet in python client:
```
ch_conf = {
    'host': CH_HOST,
    'port': CH_PORT,
    'username': CH_USER,
    'password': CH_PASSWORD,
    'rename_response_column': 'REMOVE_PREFIX',
}

-----------------------------------------------------------------------
    105                     name = name[3:]
    106                 settings[name] = value
--> 107     return HttpClient(interface, host, port, username, password, database, settings=settings, **kwargs)
    108 raise ProgrammingError(f'Unrecognized client type {interface}')

TypeError: __init__() got an unexpected keyword argument 'rename_response_column'    105                     name = name[3:]
    106                 settings[name] = value
--> 107     return HttpClient(interface, host, port, username, password, database, settings=settings, **kwargs)
    108 raise ProgrammingError(f'Unrecognized client type {interface}')

TypeError: __init__() got an unexpected keyword argument 'rename_response_column'
```

How I can rename column with Alias without using ` X.name as name`?

Example:
```
with 
	x as (select 1 as a, 2 as id),
	y as (select 1 as a, 3 as id)
select
	y.a,
	y.id
from 
	x
	join y using (a)
```

will return:
```
y.a|y.id|
---+----+
 1|   3|
```

but the expected will be:
```
 a| id|
---+----+
 1|   3|
```

joe-clickhouse:
## Summary
Adds a client option called `rename_response_column` that will automatically rename columns according to one of the chosen renaming schemes. Options are:

Option | Original Column Name | New Column Name |
------- | ---------------------- | ------------------- |
 None | `a.b.c d_e`                    | `a.b.c d_e`              |
remove_prefix | `a.b.c d_e`                    | `c d_e`              |
 to_camelcase | `a.b.c d_e`                    | `a.b.cDE`              |
 to_camelcase_without_prefix | `a.b.c d_e`                    | `cDE`              |
 to_underscore | `a.b.c d_e`                    | `a.b.c_d_e`              |
 to_underscore_without_prefix | `a.b.c d_e`                    | `c_d_e`              |

Follows closely the Java implementation from https://github.com/ClickHouse/clickhouse-java/pull/923.

Closes #228 

## Checklist


clickhouse_connect/driver/httpclient.py
```diff
@@ -283,6 +291,12 @@ def _query_with_context(self, context: QueryContext) -> QueryResult:
         byte_source = RespBuffCls(ResponseSource(response))  # pylint: disable=not-callable
         context.set_response_tz(self._check_tz_change(response.headers.get('X-ClickHouse-Timezone')))
         query_result = self._transform.parse_response(byte_source, context)
+        if self._column_renamer:
```
genzgd, line 294: It would be nice to move this functionality into the _transform.parse_response method or the query_result object.   The client doesn't need to know how to modify query_result data, and the _column_renamer seems like it belongs more to the context.  The only annoyance is that the "LIMIT 0" hack is currently in the client because ClickHouse Native won't return column names :(

genzgd (approved):

mshustov:
@joe-clickhouse following our discussion the lsat week - do you want to revert the change?

Issue #549, opened 2025-09-05
it should be easy to set a role for a query or session


svix-jbrown:
**Is your feature request related to a problem? Please describe.**

There appears to be no way to set a specific role for a query or group of queries. It's not legal to run a query like `SET ROLE foo;SELECT bar`, and wouldn't really make sense anyway.

**Describe the solution you'd like**
The HTTP protocol (which clickhouse-connect appears to use) supports [setting a role with query parameters](https://clickhouse.com/docs/interfaces/http#setting-role-with-query-parameters), so ideally we could just do that.

I believe this is a very small PR.

PR #550, opened 2025-09-05
"role" is a valid transport setting

svix-jbrown:
## Summary
This makes it possible to set a role for a given query. It's still not possible to set *multiple* roles because the way this is expressed in the protocol is by *repeating* the `role=` query parameter, which this application doesn't know how to do. 

A better full-featured solution would be to add a `role=` kwarg that takes a set of roles and builds the query parameters correctly. That's a lot of work, though, so I wanted to offer this quick/hacky workaround to unblock myself for now.

#549

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ The supported method of passing ClickHouse server settings is to prefix such arg
 
 ## UNRELEASED
 
+- Allow passing `role` as a field in the `settings` keyword argument to set a role for a specific query
+
 ## 0.9.1, 2025-09-16
 
 - Fix problem with typing that forced numpy to be installed.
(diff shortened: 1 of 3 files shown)


joe-clickhouse:
I don't have any objections to adding "role" to the `valid_transport_settings`. Seems easy then to be able to use it via "settings". Also helps prevent kwarg bloat on the client itself. The only thing I'd suggest is adding a test to verify that it works. E.g. I'd expect something like this should then work as expected.

```python
def test_role_setting_enforces_max_rows(test_client: Client):
    role_limited = "limit_rows_role"
    user_limited = "limit_rows_user"

    temp_user_password = ""
    test_client.command(f"CREATE ROLE IF NOT EXISTS {role_limited} SETTINGS max_rows_to_read = 5")
    test_client.command(f"GRANT SELECT ON system.numbers TO {user_limited}")
    test_client.command(f"CREATE USER IF NOT EXISTS {user_limited} IDENTIFIED WITH no_password")
    test_client.command(f"GRANT {role_limited} TO {user_limited}")
    role_client = create_client(
            host=test_client.server_host_name,
            port=test_client.url.rsplit(":", 1)[-1].split("/")[0],
            username=user_limited,
            password=temp_user_password,
            settings={"role": role_limited}
        )

    with pytest.raises(DatabaseError, match="TOO_MANY_ROWS"):
        role_client.query("SELECT number FROM system.numbers LIMIT 10")

    res = role_client.query("SELECT number FROM system.numbers LIMIT 3")
    assert len(res.result_rows) == 3

    role_client.close()

```

svix-jbrown:
I added a different test, since that one felt a bit mysterious to debug if it ever broke. LMK what you think.

joe-clickhouse (approved):
LGTM

PR #569, opened 2025-10-07
Always apply the timezone to datetime and datetime64 types

### LINKED ISSUE #566: No timezone localisation for results when query_tz == pytz.UTC
### Describe the bug
When a `query_df` is called with `query_tz= "UTC"` or `query_tz=pytz.UTC` the result Timestamp columns are timezone naive weither they have  `DateTime64` or `DateTime` columns as origin.

After analysing the code, it turns out that this seems intentional as inside `active_tz` the returned timezone is explicitly `None` when `active_tz = pytz.UTC`. I'm not sure why, and if there is a good reason it should be added in the documentation. 

 https://github.com/ClickHouse/clickhouse-connect/blob/main/clickhouse_connect/driver/query.py#L171

```python
    def active_tz(self, datatype_tz: Optional[tzinfo]):
        if self.column_tz:
            active_tz = self.column_tz
        elif datatype_tz:
            active_tz = datatype_tz
        elif self.query_tz:
            active_tz = self.query_tz
        elif self.response_tz:
            active_tz = self.response_tz
        elif self.apply_server_tz:
            active_tz = self.server_tz
        else:
            active_tz = tzutil.local_tz
        if active_tz == pytz.UTC:
            return None
        return active_tz
```

But even if this method is fixed and does return `UTC` when `query_tz` set to `UTC` (and this solves the issue for `DateTime`)  the `DateTime64._read_column_binary` method explicitly ignores setting the timezone if `active_tz = pytz.UTC`
https://github.com/ClickHouse/clickhouse-connect/blob/main/clickhouse_connect/datatypes/temporal.py#L246

```python
    def _read_column_binary(self, source: ByteSource, num_rows: int, ctx: QueryContext, _read_state: Any) -> Sequence:
        if self.read_format(ctx) == 'int':
            return source.read_array('q', num_rows)
        active_tz = ctx.active_tz(self.tzinfo)
        if ctx.use_numpy:
            np_array = numpy_conv.read_numpy_array(source, self.np_type, num_rows)
            if ctx.as_pandas and active_tz and active_tz != pytz.UTC:
                return pd.DatetimeIndex(np_array, tz='UTC').tz_convert(active_tz)
            return np_array
        column = source.read_array('q', num_rows)
        if active_tz and active_tz != pytz.UTC:
            return self._read_binary_tz(column, active_tz)
        return self._read_binary_naive(column)
```

### Steps to reproduce
1. See test in Code example

### Expected behaviour
When `query_tz` is set to UTC, the response's date time columns should be localised to UTC. 

### Code example
This test can executed by being appended to  `tests/test_pandas.py` file.

```python
@pytest.mark.parametrize('tz_to_test', [
        None,  # No timezone - should return naive datetime => succeeds
        'Europe/Madrid', # Timezone - should return timezone aware datetime => succeeds
        'UTC', # Timezone - should return timezone aware datetime => fails
    ])
def test_pandas_timezone(test_client: Client, table_context: Callable, tz_to_test):
    if not test_client.min_version('20'):
        pytest.skip(f'DateTime64 not supported in this server version {test_client.server_version}')

    nano_timestamp = pd.Timestamp(1992, 11, 6, 12, 50, 40, 7420, 44)
    with table_context('test_pandas_timezone', ['key String',
                                            "millis DateTime",
                                            "nanos DateTime64(9)"]):
        df = pd.DataFrame([['key1', nano_timestamp, nano_timestamp]], columns=['key', 'millis', 'nanos'])
        test_client.insert_df('test_pandas_timezone', df)
        result_df = test_client.query_df('SELECT * FROM test_pandas_timezone', query_tz=tz_to_test)
        if tz_to_test is None:
            assert result_df.iloc[0]['nanos'].tzinfo is None
            assert result_df.iloc[0]['millis'].tzinfo is None
        else:
            assert result_df.iloc[0]['millis'].tzinfo.zone == tz_to_test
            assert result_df.iloc[0]['nanos'].tzinfo.zone == tz_to_test
```

This test fails for the UTC case with error log: 
```
tests/integration_tests/test_pandas.py:202 (test_pandas_timezone[UTC])
test_client = <clickhouse_connect.driver.httpclient.HttpClient object at 0x7ebedc8f3e00>
table_context = <function table_context_fixture.<locals>.context at 0x7ebedc806f20>
tz_to_test = 'UTC'

    @pytest.mark.parametrize('tz_to_test', [
        None,  # No timezone - should return naive datetime => succeeds
        'Europe/Madrid', # Timezone - should return timezone aware datetime => succeeds
        'UTC', # Timezone - should return timezone aware datetime => fails
    ])
    def test_pandas_timezone(test_client: Client, table_context: Callable, tz_to_test):
        if not test_client.min_version('20'):
            pytest.skip(f'DateTime64 not supported in this server version {test_client.server_version}')
    
        nano_timestamp = pd.Timestamp(1992, 11, 6, 12, 50, 40, 7420, 44)
        with table_context('test_pandas_timezone', ['key String',
                                                "millis DateTime",
                                                "nanos DateTime64(9)"]):
            df = pd.DataFrame([['key1', nano_timestamp, nano_timestamp]], columns=['key', 'millis', 'nanos'])
            test_client.insert_df('test_pandas_timezone', df)
            result_df = test_client.query_df('SELECT * FROM test_pandas_timezone', query_tz=tz_to_test)
            if tz_to_test is None:
                assert result_df.iloc[0]['nanos'].tzinfo is None
                assert result_df.iloc[0]['millis'].tzinfo is None
            else:
                assert result_df.iloc[0]['millis'].tzinfo.zone == tz_to_test
>               assert result_df.iloc[0]['nanos'].tzinfo.zone == tz_to_test
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E               AttributeError: 'NoneType' object has no attribute 'zone'

tests/integration_tests/test_pandas.py:224: AttributeError
```

### Proposed solution
Modify `active_tz` function to return the `query_tz` if explicitly set by the caller as it should logically prime over other tzs.
Remove the `and active_tz != pytz.UTC` condition from the `DateTime64._read_column_binary` method.
After doing the tests on my side this solution solves this problem and does not break any other tests, if this is a genuine issue and not just an undocumented edge case for some resons that escapes me I can do a PR. 

* clickhouse-connect version: 0.9.2
* Python version: 3.13
* Operating system: Ubuntu 22.04.5 LTS

prokofyevDmitry:
## Summary
Modified the `DateTime64._read_column_binary` and `QueryContext.active_tz` code to always apply UTC instead of ignoring it as this would allow to eliminate the side effect of ignoring `query_tz` param passed to `query` method when it's equal to UTC. 

It also speed up the decoding of datetimes (see benchmarks in  issue : https://github.com/ClickHouse/clickhouse-connect/issues/566 )

## Question relative to failing tests : 
This modification does break tests because in many cases they expect naive datetimes that have now become tz-aware.

While in some cases they can be easily fixed by updating the values, for some this simple solution seems wrong:

For example test `tests/integration_tests/test_params.py`: 
```python
    dt_params = [datetime(2023, 6, 1, 7, 40, 2), datetime(2023, 8, 17, 20, 0, 10)]
    result = test_client.query('SELECT {l:Array(DateTime)}', parameters={'l': dt_params}).first_row
    assert dt_params == result[0]
```
fails as `result[0]` is now tz-aware because the QueryContext default configuration applies the server timezone. 

If the maintainers are OK with this behaviour I can easily fix all the tests by adding the expected tz to the result, if not, additional discussion is needed for those cases ( i.e : do we change the query context used in those tests, adding extra documentation about the fact that all the date-times are always returned with tz or something else). 

## Question relative to benchmarking
The benchmarks that validated the feasibility of this approach have been added into `test/benchmark` as we use the `pytest-benchmark` suite. I would appreciate the advice of the maintainers on whether this placement is the right  or if we should drop the benchmarks all together. 

## Checklist

Closes #566

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@ The supported method of passing ClickHouse server settings is to prefix such arg
 
 ## UNRELEASED
 
+### Improvements
+- Added `utc_tz_aware` parameter to client and query methods to opt in to returning timezone-aware UTC objects for DateTime/DateTime64 columns. Default behavior remains the same and returns tz naive objects for backward compatibility. Note: this parameter will likely be removed and only return tz-aware dts in some future release.
+
 ## 0.9.2, 2025-09-25
 
 - Updated `python_requires` to drop Python 3.8 and advertise support for 3.9–3.13


joe-clickhouse:
Thanks @prokofyevDmitry! Appreciate this. I'll take a look at this this week and provide some guidance on the tests.

joe-clickhouse:
Hey @prokofyevDmitry, thanks for the excellent work on this! I've made a few adjustments to the implementation based on our library's compatibility requirements and roadmap. Here's what changed I and why:

1. Benchmark tests removed
The benchmark work you did was super helpful and validated that the performance concerns around timezone-aware datetimes are no longer relevant. That said, the benchmark tests themselves are pretty niche and don't provide ongoing value in our test suite. They served their purpose as proof-of-concept so I've removed them but the findings are noted and appreciated. And for posterity the results can still be viewed in the [original issue](https://github.com/ClickHouse/clickhouse-connect/issues/566#issuecomment-3366186785).

2. Default behavior
I've flipped the default behavior from what you originally implemented. Instead of defaulting to timezone-aware UTC datetimes, the library will now continue to return naive UTC datetimes by default (`utc_tz_aware=False`). I don't want to break existing code that expects naive datetimes, even if it's technically the less correct behavior (Python best practices [strongly recommend](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow) always using timezone-aware datetimes). Users can opt-in to the better behavior by `utc_tz_aware=True` at the client level or per-query. As we approach the 1.0 release, I'm planning to flip this default as part of a batch of breaking changes. But for now, I'd prefer maintaining compatibility with existing code.

3. Tests
Since we're preserving backward compatibility by default, the existing tests continue to pass without modification. I did add a new integration test (test_utc_tz_aware in test_timezones.py) that explicitly tests both behaviors

4. Docs
I added the `utc_tz_aware` parameter to both `create_client()` and `create_async_client()` docstrings so users know it exists and what it does.

In summary, your implementation was solid and the approach is right. The only real change is flipping the default to maintain backward compatibility for now. Great work on this, and thanks for the thorough investigation with the benchmarks!

prokofyevDmitry:
Thank you for the detailed feedback !

PR #578, opened 2025-10-30
Add executor parameter to AsyncClient

meirdev:
## Summary
Currently `AsyncClient` creates a `ThreadPoolExecutor` in its constructor, I suggest allowing it to pass a `ThreadPoolExecutor` to solve two things:

1. In case more than one `AsyncClient` is needed, there is no need to create a `ThreadPoolExecutor` for each connection.
2. If the user called the `run_in_executor` function in his program, there is already a `ThreadPoolExecutor` that can be used and there is no need to create a new one.

## Checklist
- [ X] A human-readable description of the changes was provided to include in CHANGELOG


wiese-m:
Allowing passing `executor_threads` parameter along with `executor` parameter feels redundant to me. Maybe it would be a good idea to leave decision for the end user to just pass `executor: Optional[Executor] = None` where `Executor` is from `concurrent.futures`? And then in `None` case we set default behavior to create `ThreadPoolExecutor()` which already mimics `max_workers` parameter from standard library.

clickhouse_connect/driver/asyncclient.py
```diff
@@ -22,13 +25,22 @@ class AsyncClient:
...
         self.client = client
         if executor_threads == 0:
             executor_threads = min(32, (os.cpu_count() or 1) + 4)  # Mimic the default behavior
-        self.executor = ThreadPoolExecutor(max_workers=executor_threads)
+        if executor is NEW_THREAD_POOL_EXECUTOR:
+            self.new_executor = True
+            self.executor = ThreadPoolExecutor(max_workers=executor_threads)
+        else:
+            self.new_executor = False
+            self.executor = executor
```
joe-clickhouse: We hit this branch if a user passed in their own executor. However, if they also supplied `executor_threads` it could be a little confusing as that gets silently ignored since the executor already brings along its own `max_executors` from when it was originally configured. Perhaps a simple debug logging statement `if executor_threads != 0:` add a note about something like "executor_threads is ignored when a custom executor is provided".

joe-clickhouse (commented):
Hi @meirdev, thanks! Overall looks good and will be useful for when users want to share a single `ThreadPoolExecutor` across multiple `AsyncClient` instances (e.g. to limit total resource usage) or when they want to use the app's existing executor for easier thread pool management. As for the PR, I think we'll just want to make the type hints more explicit and prevent allowing `executor_threads` to be `None`.

PR #582, opened 2025-10-31
fix tls test path in contribution docs

wiese-m:
## Summary
Fixed TLS test path in contribution docs.


wiese-m:
Duplicate of #583

PR #583, opened 2025-10-31
fix version import

### LINKED ISSUE #586: Inconsistent type hint in async client definition
In `async def create_async_client` ([here](https://github.com/ClickHouse/clickhouse-connect/blob/def9a93ae840f9085b2da7bdb3cc1dfbfd0087b9/clickhouse_connect/driver/__init__.py#L149)) we allow `executor_threads` to be optional. However, the `AsyncCleint` class definition [here](https://github.com/ClickHouse/clickhouse-connect/blob/def9a93ae840f9085b2da7bdb3cc1dfbfd0087b9/clickhouse_connect/driver/asyncclient.py#L25) we force it to be an `int`. If the user does not supply `executor_threads` in `get_async_client`, default of `None` is passed to `AsyncCleint.__init__` which it accepts. `if executor_threads == 0:` fails because it's `None`, and we end up doing `self.executor = ThreadPoolExecutor(max_workers=None)`. It just so happens that the behavior of creating a `ThreadPoolExecutor` instance when `max_workers` is `None` is the same as what we do with a value of 0: `executor_threads = min(32, (os.cpu_count() or 1) + 4)` so we're seeing consistent behavior on accident. We just need to fix the type hint and default behavior in `get_async_client` for `executor_threads`.

wiese-m:
## Summary
Fixed version import in `playtest.py` & added missing return type in `version()`

Closes #586
Closes #582

### DIFF
--- a/playtest.py
+++ b/playtest.py
@@ -1,9 +1,8 @@
 import clickhouse_connect
-from clickhouse_connect.common import version
 
 
 def main():
-    print(f'\nClickHouse Connect installed version: {version()}')
+    print(f'\nClickHouse Connect installed version: {clickhouse_connect.version()}')
     client = clickhouse_connect.get_client(host='play.clickhouse.com',
                                            username='play',
                                            password='clickhouse',


playtest.py
```diff
@@ -1,8 +1,9 @@
 import clickhouse_connect
+from clickhouse_connect.common import version
```
joe-clickhouse: i'd rather not directly expose functions from `common` by importing it as that's more of an collection of internal helper functions, not really part of the core API so I'd prefer leaving this one as it was.

joe-clickhouse (commented):
Thanks @wiese-m. Appreciate the PR. See my comment below about `common`. As for the return type, that's a good call. Though there are some other functions in `common` that could benefit from return types as well. So if you're up for it, I'd suggest adding correct return types to the rest of the functions in common if they're missing them!

I also recently found a default/typing-related bug tracked in https://github.com/ClickHouse/clickhouse-connect/issues/586. All that's required there is changing the type of `executor_threads` parameter in `create_async_client` from `Optional[int]` to just a plain `int` and then change the default value from `None` to `0`. 

I see you've opened #582 as well (thanks!). So if you wouldn't mind, I suggest combing all this into a single PR that:
1. Adds return types to all the functions missing them in common.py
2. Fixes the default value and type of `executor_threads` from #586
3.  Fixes the test path you issue you found in #582
4. Reverts the `version` import in palaytest.py and uses the public entry point of `clickhosue_connect.version()`

Let me know if you have any questions and thanks again!

playtest.py
wiese-m: Sure, I reverted my changes made to `playtest.py`. My point was that in current state it does not work due to no `version` variable or function is exposed in public entry point, or am I wrong?

For me it raises exception:
`AttributeError: module 'clickhouse_connect' has no attribute 'version'`

joe-clickhouse: Thanks! Ok, I didn't realize this was broken. I'll open up a separate issue for that. Plus the canonical way in python to expose a version is like

```python
import clickhouse_connect
clickhouse_connect.__version__
```

which doesn't work in the current state and I've been meaning to get this working anyway, so I'll track that separately. Thanks!

joe-clickhouse (approved):
Looks good! Thanks for the help.

PR #584, opened 2025-10-31
Support for array join and final modifier in sqlalchemy

### LINKED ISSUE #579: Support for array join and final in sqlalchemy
1. There is support for several join types, but not yet `ARRAY JOIN`
2. We should support `FINAL` for replacing table engines

joe-clickhouse:
## Summary
1. Fixes what appears to be a bug in `create_insert_context` function that missed passing `settings` causing queries that rely on consistent settings across an operation to potentially fail.

2. Adds support to the SQLAlchemy core API for:
- `[LEFT] ARRAY JOIN`s e.g.
- `FINAL` modifier for use with `Replacing*` tables

## Examples
Assume this is our table:
| id | name | tags |
|----|---------|--------------------------------|
| 1 | Alice | ['python', 'sql', 'clickhouse'] |
| 2 | Bob | ['java', 'sql'] |
| 3 | Joe | ['python', 'javascript'] |
| 4 | Charlie | [] |

### `ARRAY JOIN`
Now you can build a query like
```python
query = (
    select(test_table.c.id, test_table.c.name, test_table.c.tags)
    .select_from(array_join(test_table, test_table.c.tags))
    .order_by(test_table.c.id)
    .order_by(test_table.c.tags)
)
```
This will result in:

| id | name | tag |
|----|-------|------------|
| 1 | Alice | "clickhouse" |
| 1 | Alice | "python" |
| 1 | Alice | "sql" |
| 2 | Bob | "java" |
| 2 | Bob | "sql" |
| 3 | Joe | "javascript" |
| 3 | Joe | "python" |

### `LEFT ARRAY JOIN`
Similarly, a `LEFT ARRAY JOIN`:
```python
query = (
    select(
        test_table.c.id,
        test_table.c.name,
        literal_column("tag"),  # Needed when using alias
    )
    .select_from(array_join(test_table, test_table.c.tags, alias="tag", is_left=True))
    .order_by(test_table.c.id)
    .order_by(literal_column("tag"))
)
```
This will result in:
| id | name | tag |
|----|-------|------------|
| 1 | Alice | "clickhouse" |
| 1 | Alice | "python" |
| 1 | Alice | "sql" |
| 2 | Bob | "java" |
| 2 | Bob | "sql" |
| 3 | Joe | "javascript" |
| 3 | Joe | "python" |
| 4 | Charlie | "" |

### `FINAL` modifier
Assume we have this table in a `ReplacingMergeTree` table
| id | name | value |
|----|-------|-------|
| 1 | Alice | 100 |
| 1 | Alice | 200 |
| 2 | Bob | 300 |

Now you can use the `.final()` method to ensure `FINAL` is included in the generated SQL, forcing ClickHouse to fully merge the data at query time. Use it like this:

```python
query = select(test_table).final().order_by(test_table.c.id)
```
This will generate SQL like:
```sql
SELECT id, name, value FROM test_table FINAL ORDER BY id
```

## Checklist

Closes #579

### DIFF
--- a/.github/workflows/on_push.yml
+++ b/.github/workflows/on_push.yml
@@ -64,12 +64,12 @@ jobs:
           - '3.11'
           - '3.12'
           - '3.13'
+          - '3.14'
         clickhouse-version:
           - '25.3' # LTS
-          - '25.6' # Stable
-          - '25.7' # Stable
           - '25.8' # LTS
           - '25.9' # Stable
+          - '25.10' # Stable
           - latest
 
     name: Local Tests Py=${{ matrix.python-version }} CH=${{ matrix.clickhouse-version }}
@@ -208,6 +208,7 @@ jobs:
           - '3.11'
           - '3.12'
           - '3.13'
+          - '3.14'
 
     steps:
       - name: Checkout
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -11,19 +11,18 @@ on:
           - test
           - release
           - build
-env:
-  CIBW_SKIP: 'cp36-* cp37-* cp38-* pp37-* pp38-*'
 
 jobs:
   build_x86_manylinux_wheels:
     name: Build x86 manylinux wheels on Linux
     runs-on: ubuntu-latest
     env:
-      CIBW_SKIP: 'cp36-* cp37-* cp38-* pp37-* pp38-* *-musllinux*'
+      CIBW_SKIP: '*-musllinux* pp38-* pp311-* cp38-* cp314t-*'
+      CIBW_ENABLE: 'pypy pypy-eol'
     steps:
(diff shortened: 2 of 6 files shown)


clickhouse_connect/driver/client.py
```diff
@@ -961,7 +961,7 @@ def create_insert_context(self,
                 full_table = quote_identifier(table)
         column_defs = []
         if column_types is None and column_type_names is None:
-            describe_result = self.query(f'DESCRIBE TABLE {full_table}')
+            describe_result = self.query(f'DESCRIBE TABLE {full_table}', settings=settings)
```
joe-clickhouse: @genzgd I was trying to track down some intermittent failures in cloud tests involving temp tables tables and think it might be related to this. Just want to make sure `settings` was not deliberately left out here.

PR #597, opened 2025-11-26
add feature to create tables from pyarrow objects

### LINKED ISSUE #588: Ability to create table from pyarrow objects
**Is your feature request related to a problem? Please describe.**
Currently there seems to be no way to create ClickHouse tables from any of the pyarrow objects (`Table`, `RecordBatch` ... ) 

**Describe the solution you'd like**
the implementation is available for inserting pyarrow objects into existing tables using `insert_arrow`, it'll be awesome if an overload for [create_table](https://github.com/ClickHouse/clickhouse-connect/blob/3e190724b2d39d956f2f6f0215f343c83a103c2c/clickhouse_connect/driver/ddl.py#L23) was made available which accepts a pyarrow `columns: pyarrow.Schema` instead of `columns: Sequence[TableColumnDef]`

```
def create_table(table_name: str, arrow_schema: pyarrow.Schema, engine: str, engine_params: dict):
    ...
```

alternatively, a utility can be added that allows conversion of a `pyarrow.Schema` to a `Sequence[TableColumnDef]`

**Describe alternatives you've considered**
The only way i see is to do manual mapping at the moment which may not be 100% correct.

**Additional context**
Another way can be to add a `create_if_not_exists: bool` param to `insert_arrow` itself which simulates a CTAS like:

```sql
CREATE TABLE {tbl_name} AS ( SELECT * from {arrow_object})
```

akkik04:
## Summary
This PR implements the helper requested in [#588](https://github.com/ClickHouse/clickhouse-connect/issues/588) for creating ClickHouse tables from PyArrow schemas.

- Adds a new utility `arrow_schema_to_column_defs(schema: pa.Schema) -> list[TableColumnDef]` that converts a `pyarrow.Schema` into `TableColumnDef` instances.
- Adds `create_table_from_arrow_schema(table_name, schema, engine, engine_params)` as a convenience wrapper that reuses the existing `create_table` helper.
- Supports core scalar Arrow types and maps them to ClickHouse types:
  - `pa.int8/16/32/64` → `Int8/16/32/64`
  - `pa.uint8/16/32/64` → `UInt8/16/32/64`
  - `pa.float16/float32` → `Float32`
  - `pa.float64` → `Float64`
  - `pa.string()/pa.large_string()` → `String`
  - `pa.bool_()` → `Bool`
- For other Arrow types, the helper raises `TypeError` so callers are explicitly aware that automatic mapping is not yet implemented.

This allows patterns like:

``` python
arrow_table = pa.table(...)
col_defs = arrow_schema_to_column_defs(arrow_table.schema)
ddl = create_table("my_table", col_defs, "MergeTree", {"ORDER BY": "id"})
client.command(ddl)
client.insert_arrow("my_table", arrow_table)
```

## Checklist: 

Closes #588


akkik04:
addressed some changes, can i get some eyes on it whenever y'all get a chance 🙌
cc: @joe-clickhouse @mshustov

clickhouse_connect/driver/ddl.py
```diff
@@ -26,3 +31,108 @@ def create_table(table_name: str, columns: Sequence[TableColumnDef], engine: str
...
+    """
+    Best-effort mapping from common PyArrow types to ClickHouse type names.
+
+    Covers core scalar types. For anything unknown, we raise so the
+    caller is aware that the automatic mapping is not implemented for that Arrow type.
+    """
+    if pa is None:
+        raise ImportError(
+            "PyArrow is required, but it is not installed."
+        )
```
joe-clickhouse: There's actually a utility in `driver/options.py` that'll do this for you. So you can replace the try/except above on lines 5-8 with `from clickhouse_connect.driver.options import check_arrow` and then here in the `_arrow_type_to_ch` function, just replace the `if pa is None` check/raise with `pa = check_arrow()`

joe-clickhouse (commented):
Thanks @akkik04! In general I think this looks pretty good. One thing that did cross my mind during review that we'll need to discuss/work through is nullable behavior. Arrow fields are nullable by default. However, this implementation creates non-nullalbe columns in ClickHouse. I wouldn't want to just automatically wrap everything in `Nullable()` since that does have performance implications. There are a couple of ways to approach this depending on what the expected workflow here is.

The tests just build up a schema from scratch (no data, just metadata) and use that to convert to `TableColumnDef` objects. This is fine for unit tests but I'm going to assume that the user will most likely have an actual arrow table with data already (if you guys envision this differently, let me know) otherwise, I'd argue they should just build the ClickHouse table in the traditional way with raw SQL in a `client.command()` statement.

Assuming they do have a table with data, we can go one of two ways:

1. null checks on arrow columns should be super cheap so we can check each column:
```python
for column in arrow_table.columns:
    has_nulls = column.null_count > 0
```
and if no nulls are found, then everything is fine as it was. If nulls are found, then wrap the type name in `Nullable()` before creating the `TableColumnDef`
```python
ch_type_name = f"Nullable({ch_type_name})"
```
2. take an 'optimistic non-null' approach and create non-nullable columns by default. If the user tries to insert an arrow table with nulls, they'll get a clear ClickHouse error indicating which column has nulls, and they can adjust the DDL accordingly. This keeps the common case (no nulls) performant while still catching issues immediately. Note that this approach assumes the nullability characteristics of the initial data are representative of future inserts. Users can always manually adjust the DDL if their data patterns change.

I'm inclined at this point to take approach 2 because it's simpler and optimizes for the common case (no nulls). The error will be immediate and actionable.

clickhouse_connect/driver/ddl.py
akkik04: This is nice and convenient, will work it in.

tests/integration_tests/test_pyarrow_ddl.py
```diff
@@ -0,0 +1,138 @@
...
+    )
+
+    ddl_wrapper = create_table_from_arrow_schema(
+        table_name="arrow_compare_test",
+        schema=schema,
+        engine="MergeTree",
+        engine_params={"ORDER BY": "id"},
+    )
+
+    assert ddl_manual == ddl_wrapper
```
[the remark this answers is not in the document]
akkik04: Agreed, will work this in as well.

akkik04:
> Thanks @akkik04! In general I think this looks pretty good. One thing that did cross my mind during review that we'll need to discuss/work through is nullable behavior. Arrow fields are nullable by default. However, this implementation creates non-nullalbe columns in ClickHouse. I wouldn't want to just automatically wrap everything in `Nullable()` since that does have performance implications. There are a couple of ways to approach this depending on what the expected workflow here is.
> 
> The tests just build up a schema from scratch (no data, just metadata) and use that to convert to `TableColumnDef` objects. This is fine for unit tests but I'm going to assume that the user will most likely have an actual arrow table with data already (if you guys envision this differently, let me know) otherwise, I'd argue they should just build the ClickHouse table in the traditional way with raw SQL in a `client.command()` statement.
> 
> Assuming they do have a table with data, we can go one of two ways:
> 
> 1. null checks on arrow columns should be super cheap so we can check each column:
> 
> ```python
> for column in arrow_table.columns:
>     has_nulls = column.null_count > 0
> ```
> 
> and if no nulls are found, then everything is fine as it was. If nulls are found, then wrap the type name in `Nullable()` before creating the `TableColumnDef`
> 
> ```python
> ch_type_name = f"Nullable({ch_type_name})"
> ```
> 
> 2. take an 'optimistic non-null' approach and create non-nullable columns by default. If the user tries to insert an arrow table with nulls, they'll get a clear ClickHouse error indicating which column has nulls, and they can adjust the DDL accordingly. This keeps the common case (no nulls) performant while still catching issues immediately. Note that this approach assumes the nullability characteristics of the initial data are representative of future inserts. Users can always manually adjust the DDL if their data patterns change.
> 
> I'm inclined at this point to take approach 2 because it's simpler and optimizes for the common case (no nulls). The error will be immediate and actionable.

Thanks for calling out the nullable behaviour, that makes sense.

Right now, the helper is effectively following your option (2): it always generates non-nullable ClickHouse types and doesn't infer `Nullable` from the PyArrow schema or data. If a user inserts a PyArrow table with nulls, ClickHouse will raise an error and they can adjust the DDL (e.g., wrap specific columns in `Nullable(...)`.

I don't have to make any functional changes to accommodate this chosen/desired behaviour, however, I'll update the docstring to explicitly document this "optimistic non-null" behaviour so the intent is clear.

clickhouse_connect/driver/ddl.py
```diff
@@ -26,3 +27,110 @@ def create_table(table_name: str, columns: Sequence[TableColumnDef], engine: str
...
+    """
+    Convert a PyArrow Schema into a list of TableColumnDef objects.
+    
+    This helper uses an *optimistic non-null* strategy: it always produces
+    non-nullable ClickHouse types, even though Arrow fields are nullable by
+    default.
+
+    If the user later inserts an Arrow table that contains nulls, ClickHouse
+    will raise an error for that insert, and the user can adjust the DDL
+    (e.g. wrap specific types in Nullable(...)) to match their data.
```
joe-clickhouse: > If the user later inserts an Arrow table that contains nulls, ClickHouse will raise an error for that insert

Did some testing and it looks like I misspoke earlier. My apologies. It turns out this is not the case when inserting via `insert_arrow()`. Client-side validation is done when using standard `insert()` to prevent `None`s from getting into non-nullable columns but during arrow insert, we don't appear to perform that kind of validation. We simply use pyarrow's `RecordBatchFileWriter` which will allow nulls to be handed to the server which then just uses assigned default values, e.g. `0`'s for `int` types, `""` for `string`, etc. because the server's `input_format_null_as_default` [setting](https://clickhouse.com/docs/operations/settings/formats#input_format_null_as_default) defaults to 1. This has nothing to do with your PR/implementation as this is a pre-existing condition, but worth noting anyway.

We should change the docstring.

```suggestion
    Note that if the user inserts a table with nulls into a non-Nullable column,
    ClickHouse will silently convert those nulls to default values due to the default
    server setting input_format_null_as_default=1.
```

I'll deal with the validation inconsistency separately.

joe-clickhouse (commented):
Thanks for the fixes! Only thing you'll need to do now is run `pylint tests` and `pylint clickhouse_connect` and address those issues as the workflow won't even run until those issues are addressed.

clickhouse_connect/driver/ddl.py
akkik04: Got it, thanks for clarifying the `insert_arrow` behavior. I’ve updated the docstring with your suggested text and pushed the change.

joe-clickhouse (approved):
Looks good! Thanks for the contribution.

PR #618, opened 2026-01-16
Bugfix: InsertContext data not reset on insert failure

### LINKED ISSUE #616: InsertContext not reset on failed insert, causing reuse to fail after exceptions if data passed separately
### Describe the bug

`InsertContext` is documented and tested as reusable, but its internal `data` field is only reset on successful inserts. When an exception occurs (for example, due to an IO/network error or a server-side failure), `context.data` can remain set. This breaks subsequent reuse of the same `InsertContext` when the caller supplies `data=` to `insert()` (rather than setting `context.data` directly), because the client detects that the context still has data and raises.

Specifically:

- `context.data` is reset to `None` only on success  
  https://github.com/ClickHouse/clickhouse-connect/blob/main/clickhouse_connect/driver/httpclient.py#L338

- When an exception occurs during the raw request, the reset never happens  
  https://github.com/ClickHouse/clickhouse-connect/blob/main/clickhouse_connect/driver/httpclient.py#L478

- On subsequent reuse, the client checks whether `context.data` is still set and raises an exception  
  https://github.com/ClickHouse/clickhouse-connect/blob/main/clickhouse_connect/driver/client.py#L803

Because of this, if an insert fails once, the same `InsertContext` cannot be reused via `client.insert(data=..., context=ctx)` even though reuse is explicitly supported and tested.

> **Note:** This behavior may not strictly be a bug in the implementation. It could also be a documentation gap or an unclear API contract around whether an `InsertContext` is expected to remain reusable after a failed insert, especially when mixing `context.data` and `data=` usage patterns.

---

### Steps to reproduce

```python
import clickhouse_connect

    host="localhost",
    password="pass",
)

ctx = client.create_insert_context(
    database="demo",
    table="test_contexts",
    column_names=["id"]
)

# Force a failure (any failure works; this is a simple server-side one)
ctx.table = "doesnotexist"
try:
    client.insert(context=ctx, data=[[1]])
except Exception as e:
    print("Insert failed as expected:", type(e).__name__, e)

print("After failure, ctx.data =", ctx.data)

# Reuse the same context but pass data separately:
# This fails because ctx.data was not cleared after the exception.
client.insert(data=[[2]], context=ctx)  # <-- fails due to stale ctx.data

# Workaround: explicitly set ctx.data (i.e. keep using the context as the data carrier)
ctx.data = [[2]]
client.insert(context=ctx)  # works
```

### Actual behaviour

After a failed insert, ctx.data is still populated. A subsequent call using
client.insert(data=..., context=ctx) fails immediately due to the stale ctx.data state.

### Expected behaviour

Either:

context.data should be reset to None in the error/exception handling path (to preserve the documented/tested reusability), or

Document that after a failed insert, reusing an InsertContext with data= passed separately is unsupported unless the caller manually clears ctx.data

Given that reuse is covered by tests (for example
https://github.com/ClickHouse/clickhouse-connect/blob/main/tests/integration_tests/test_contexts.py
), resetting the context on failure seems like the more intuitive behavior.

### Configuration

- clickhouse-connect version: 0.10.0
- Python version: 3.13
- Operating system: Linux

abhisheksurve45:
## Summary

Fixes #616

`InsertContext` is documented and tested as reusable, but its internal `data` field is only reset on successful inserts. When an exception occurs (for example, due to an IO/network error or a server-side failure), `context.data` can remain set. This breaks subsequent reuse of the same `InsertContext` when the caller supplies `data=` to `insert()` (rather than setting `context.data` directly), because the client detects that the context still has data and raises.

Specifically:

- `context.data` is reset to `None` only on success  
  https://github.com/ClickHouse/clickhouse-connect/blob/main/clickhouse_connect/driver/httpclient.py#L338

- When an exception occurs during the raw request, the reset never happens  
  https://github.com/ClickHouse/clickhouse-connect/blob/main/clickhouse_connect/driver/httpclient.py#L478

- On subsequent reuse, the client checks whether `context.data` is still set and raises an exception  
  https://github.com/ClickHouse/clickhouse-connect/blob/main/clickhouse_connect/driver/client.py#L803

Because of this, if an insert fails once, the same `InsertContext` cannot be reused via `client.insert(data=..., context=ctx)` even though reuse is explicitly supported and tested.

This PR resets the `context.data` in `finally` block so gets reset on both success and failure cases.

## Checklist

### DIFF
--- a/tests/integration_tests/test_contexts.py
+++ b/tests/integration_tests/test_contexts.py
@@ -1,4 +1,5 @@
 from typing import Callable
+import pytest
 
 from clickhouse_connect.driver import Client
 
@@ -27,3 +28,15 @@ def test_contexts(test_client: Client, table_context: Callable):
         insert_context.data = [[5, 'v5', 'v6'], [7, 'v7', 'v8']]
         test_client.insert(context=insert_context)
         assert test_client.command(f'SELECT count() FROM {ctx.table}') == 6
+
+def test_insert_context_data_cleared_on_failure(test_client: Client, table_context: Callable):
+    with table_context('test_contexts', ['key Int32', 'value1 String', 'value2 String']) as ctx:
+        data = [[1, "v1", "v2"], [2, "v3", "v4"]]
+        insert_context = test_client.create_insert_context(table=ctx.table, data=data)
+
+        insert_context.table = f"{ctx.table}__does_not_exist"
+
+        with pytest.raises(Exception):
+            test_client.insert(context=insert_context)
+
+        assert insert_context.data is None


joe-clickhouse:
Hi @abhisheksurve45 I haven't forgot about trying to get a release out! I'm struggling with some cloud tests at the moment but think I have a path forward so as soon as CI is sorted and working as expected, I'll get a release out for ya.

Issue #377, opened 2024-07-23
Timezone-aware `datetime` query parameters are internally cast to a wrong timezone


maxblumental:
### Describe the bug

Timezone-aware `datetime` query parameters are internally cast to a wrong timezone.

### Steps to reproduce
1. Use a database whose timezone is EST (UTC-05:00):
```sql
SELECT timezone();
-- EST
```
2. Make a `Client` with default parameters:
```python
import clickhouse_connect

client = clickhouse_connect.get_client(
    username='...',
    password='...',
    host='...',
    port=443,
    database='...',
)
```
3. Make a query with a timezone-aware datetime parameter in UTC+3:
```python
import datetime
tzinfo = datetime.timezone(offset=datetime.timedelta(hours=3)) # UTC+3
dt = datetime.datetime(2024, 7, 23, 12, 0, 0, tzinfo=tzinfo)
result = client.query('select toString(%s)', parameters=(dt,))
print(result.result_rows[0][0])
# 2024-07-23 09:00:00
```

It should have converted UTC+3 to UTC-5 and get `2024-07-23 04:00:00`.

### Where it happens

As we can see in the debugger, the function `clickhouse_connect.driver.query.py:format_query_value` is casting the parameter to UTC:
```python
    if isinstance(value, datetime):
        if value.tzinfo is not None or server_tz != pytz.UTC:
            value = value.astimezone(server_tz) # <-- here
        return f"'{value.strftime('%Y-%m-%d %H:%M:%S')}'"
```

The `format_query_value()` was called from `finalize_query()`:

```python
def finalize_query(query: str, parameters: Optional[Union[Sequence, Dict[str, Any]]],
                   server_tz: Optional[tzinfo] = None) -> str:
    while query.endswith(';'):
        query = query[:-1]
    if not parameters:
        return query
    if hasattr(parameters, 'items'):
        return query % {k: format_query_value(v, server_tz) for k, v in parameters.items()}
    return query % tuple(format_query_value(v) for v in parameters) # <-- here
```

For some reason, `server_tz` wasn't passed to `format_query_value()` and `pytz.UTC` default value was used.

#### Environment
* clickhouse-connect version: 0.7.16
* Python version: 3.9.6
* Operating system: macOS Ventura 13.5.2

#### ClickHouse server
* ClickHouse Server version: 23.8.8.21.altinitystable

genzgd:
[...]

Issue #308, opened 2024-03-01
Unexpected timezone-aware return object for UTC query


tatjanapuskarov:
### Describe the bug

When querying a `DateTime` column with the applied timezone being UTC, it is possible to get a timezone-aware datetime object as a response, even though a timezone-naive object is expected according to the [documentation](https://clickhouse.com/docs/en/integrations/python#time-zones):
> Note that if the applied timezone based on these rules is UTC, clickhouse-connect will always return a time zone naive Python datetime.datetime object. 

If `active_tz` evaluates to UTC according to [these rules](https://github.com/ClickHouse/clickhouse-connect/blob/8415839e87345677bb585820880918bd47aae718/clickhouse_connect/driver/query.py#L161C1-L172C38) and `local_tz` is not UTC, `active_tz` is set to UTC (and not `None`) because of [this check](https://github.com/ClickHouse/clickhouse-connect/blob/8415839e87345677bb585820880918bd47aae718/clickhouse_connect/driver/query.py#L175), even though `local_tz` should not be taken into account according to the `active_tz` setting precedence order. 

With `active_tz` being UTC and querying a `DateTime` column, the output is a timezone-aware object because of [this condition](https://github.com/ClickHouse/clickhouse-connect/blob/8415839e87345677bb585820880918bd47aae718/clickhouse_connect/datatypes/temporal.py#L124). For `DateTime64`, the output is a timezone-naive object because [the equivalent condition](https://github.com/ClickHouse/clickhouse-connect/blob/8415839e87345677bb585820880918bd47aae718/clickhouse_connect/datatypes/temporal.py#L174) includes an extra check.

### Expected behaviour

- The behaviour of `DateTime` and `DateTime64` types is expected to be consistent and should return a timezone-naive object regardless of whether `active_tz` is `None` or `'UTC'` (regardless of `local_tz`).
- The [documentation](https://clickhouse.com/docs/en/integrations/python#time-zones) should include usage of the local timezone.

### Code example
```python
import clickhouse_connect
import pytest
import pytz

with pytest.MonkeyPatch.context() as mp:
    mp.setattr("clickhouse_connect.driver.context.BaseQueryContext.local_tz", pytz.timezone('Pacific/Johnston')) # any non-UTC timezone 
    client = clickhouse_connect.get_client(
                host=…,
                port=…,
                username=…,
                password=…,
                database="default"
            )

   query = """
            SELECT
                datetime,
                datetime64
            FROM default.utc_bug
[...]

Issue #210, opened 2023-06-27
Timezone info is not present in the result when querying with `columns_tzs` or `query_tz`


pkit:
Simple repro:

```python
import clickhouse_connect

c = clickhouse_connect.create_client()
c.command("CREATE TABLE t1 (id UInt64, ts DateTime64(3, 'UTC') DEFAULT now64(3)) ENGINE = MergeTree() ORDER BY id")
c.insert("t1", data=[(1,)], column_names=["id"])
rows = c.query("SELECT id, ts FROM t1 ORDER BY ts", column_tzs={"ts": "UTC"}).result_set
print(rows)
rows = c.query("SELECT id, ts FROM t1 ORDER BY ts", query_tz="UTC").result_set
print(rows)
```

Produces "naive" datetime objects without any TZ info:
```
[(1, datetime.datetime(2023, 6, 27, 19, 47, 53, 536000))]
[(1, datetime.datetime(2023, 6, 27, 19, 47, 53, 536000))]
```

Version of clickhouse-connect: `0.6.4`

genzgd:
This is expected behavior.  As stated in the changelog for version 0.5.17:

```
Note if the detected timezone according to the above precedence is UTC, clickhouse-connect will always return a naive datetime object with no timezone information```

Applying a timezone is quite expensive in Python and if the user really needs a UTC timezone applied, they should do in their own application code.

pkit:
This one is not according to the python standard, see: https://docs.python.org/3/library/datetime.html#datetime.datetime.astimezone
> If self is naive, it is presumed to represent time in the system timezone.

genzgd:
In this case I think it's reasonable to prefer performance over a convention/presumption.

The most common use case for ClickHouse clients is in container applications using UTC timestamps.  Converting ClickHouse integer epoch timestamps (which is how all times are stored in ClickHouse) to Python naive datetime objects is already annoyingly expensive (numpy datetime or Pandas objects are much faster and cleaner).  Adding even more processing cost by making them UTC timezone aware when that conversion gains nothing in the majority of applications is something that I think the application developer should do consciously.

pkit:
Without convention it cannot be passed to any other library, as it will implicitly assume that internally it's a system time, and it can be pretty deep in other people code.
I.e. safest bet would be just to convert it to tz-aware anyway...

> conversion gains nothing in the majority of applications

[...]

Issue #136, opened 2023-02-22
Timezone-aware DateTime64 columns now returned missing the timezone info


georgipeev:
### Describe the bug
Timezone-aware columns in Clickhouse like `(DateTime64(6, 'America/Chicago'))` used to be returned by `query_df` in correspondingly timezone-aware DataFrame columns of type `pandas.DatetimeTZDtype(unit='ns', tz='America/Chicago')` until (at least) version 0.5.7 of clickhouse-connect. In the latest version 0.5.12 such columns are returned as timezone-unaware `numpy.datetime64` columns.

### Steps to reproduce
1. Create a table with a `(DateTime64(6, 'America/Chicago'))` column; insert some data
2. Query the table using `query_df`
3. The dtype of the column in the returned DataFrame will be `pandas.DatetimeTZDtype(unit='ns', tz='America/Chicago')` if using clickhouse-connect version 0.5.7 (i.e., timezone-aware), but `numpy.datetime64` if using clickhouse-connect version 0.5.12 (i.e., timezone-unaware).

### Expected behaviour
Returned colunm dtype is `pandas.DatetimeTZDtype(unit='ns', tz='America/Chicago')`

genzgd:
That is related to https://github.com/ClickHouse/clickhouse-connect/issues/120.

For various obscure historical reasons ClickHouse *does* return the timezone information for a DateTime64 column, but *doesn't* return that information for a DateTime column (in the particular case of HTTP using Native format) so `clickhouse-connect` was not setting timezone for any DateTime column (after the internal switch from RowBinary to Native format).  There's a workaround where the client can specific the timezone as a keyword argument (`query_tz`/`column_tzs`), but even that doesn't work for pandas currently.

I have a fix for the DateTime64 issue, which will also enable the `query_tz`/`column_tzs` workaround.  That fix will probably be released early next week.  Also, in today's release of ClickHouse, there's also a fix for the problem of not returning timezone information for DateTime columns.  However, that will only work with the most recent ClickHouse version 23.2.x and also requires the next `clickhouse-connect` release to handle the low level change.  So eventually, with the latest ClickHouse server and clickhouse-connect version, the workaround won't be necessary.

(Of course, my long standing  -- about 20 years now -- recommendation is **never** store anything related to timezones in your database, just store absolute UTC times, and wait until the final user UI to actually display the adjusted time.  Oh well.)

georgipeev:
[...]

PR #621, opened 2026-01-26
fix: handle Etc/UTC timezone equivalence for naive datetime conversion

Milias:
# Description

Hello everyone,

I noticed that when querying timestamp columns **without** timezone I would always get back timestamps **with** timezones, which breaks joins and other operations in polars when other columns do not have timezones. This should fix this problem. Below I link to several previous issues somewhat related to this issue, that did not fix it fully.

It's the first time contributing to this repo (or any other ClickHouse repos!) so please let me know if anything is not how you like it :) happy to make changes!

---

> Claude-generated from this point.

## Summary

- Fix `pytz.timezone('Etc/UTC') != pytz.UTC` comparison bug causing timezone-aware datetimes when `utc_tz_aware=False`
- Add `is_utc_timezone()` helper to properly detect UTC equivalents (Etc/UTC, GMT, Universal, etc.)
- Strip UTC timezone from Arrow format output for pandas/polars DataFrames

## Problem

When a ClickHouse server's timezone is set to `Etc/UTC`, queries return `datetime[ns, Etc/UTC]` instead of the expected `datetime[ns]` (naive) when `utc_tz_aware=False` (the default).

**Root Cause**: In Python's pytz library, `pytz.timezone('Etc/UTC') != pytz.UTC` despite being semantically equivalent. The comparison at `query.py:188` uses `active_tz == pytz.UTC`, which fails when the server timezone is `Etc/UTC`.

## Related Issues

| Issue | Title | Relevance |
|-------|-------|----------|
| [#308](https://github.com/ClickHouse/clickhouse-connect/issues/308) | Unexpected timezone-aware return object for UTC query | **Direct predecessor** - fixed DateTime vs DateTime64 inconsistency in v0.7.2 but missed the `Etc/UTC` case |
| [#377](https://github.com/ClickHouse/clickhouse-connect/issues/377) | Timezone-aware datetime parameters cast to wrong timezone | Related - `server_tz` not passed to `format_query_value()` |
| [#210](https://github.com/ClickHouse/clickhouse-connect/issues/210) | Timezone info not present with `query_tz` | Inverse issue - UTC returns naive but user wanted aware |
| [#136](https://github.com/ClickHouse/clickhouse-connect/issues/136) | DateTime64 columns missing timezone info | Breaking change in timezone handling |

### Key Finding from #308

> "When querying a DateTime column with the applied timezone being UTC, it is possible to get a timezone-aware datetime object as a response, even though a timezone-naive object is expected according to the documentation."

The fix in v0.7.2 addressed DateTime vs DateTime64 inconsistency but did **not** address the case where `pytz.timezone('Etc/UTC')` is used instead of `pytz.UTC`.

### pytz Timezone Comparison Problem

From [pandas-dev/pandas#23959](https://github.com/pandas-dev/pandas/issues/23959):
> "`pd._libs.tslibs.timezones.tz_compare('UTC', pytz.timezone('UTC'))` returns `False`"

This is a known quirk where timezone equality is defined by object identity, not semantic equivalence.

### `clickhouse_connect/driver/tzutil.py`
- Add `is_utc_timezone()` helper that checks for UTC equivalents via `zone` attribute and `tzname()` method

### `clickhouse_connect/driver/query.py` (line 188)
```python
# Before:
if active_tz == pytz.UTC and not self.utc_tz_aware:
# After:
if tzutil.is_utc_timezone(active_tz) and not self.utc_tz_aware:
```

### `clickhouse_connect/driver/binding.py` (line 128)
```python
# Before:
if value.tzinfo is not None or server_tz != pytz.UTC:
# After:
if value.tzinfo is not None or not tzutil.is_utc_timezone(server_tz):
```

### `clickhouse_connect/driver/client.py`
- Add `_strip_utc_timezone_from_arrow()` helper to cast `timestamp[us, UTC]` and `timestamp[us, Etc/UTC]` to naive `timestamp[us]`
- Apply to `query_df_arrow()` and `query_df_arrow_stream()` converters for both pandas and polars

## Test Plan

- [x] Add unit tests for `Etc/UTC` handling in `QueryContext.active_tz()`
- [x] Add integration test for polars DataFrame UTC timestamp stripping
- [x] Verify `is_utc_timezone()` correctly identifies UTC, Etc/UTC, GMT, and rejects non-UTC timezones
- [x] Verify Arrow timezone stripping preserves non-UTC timezones (e.g., America/Denver)
- [ ] Run full test suite (requires ClickHouse server for integration tests)

---

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/tests/integration_tests/test_polars.py
+++ b/tests/integration_tests/test_polars.py
@@ -121,7 +121,7 @@ def test_polars_utc_timestamp_naive(test_client: Client, table_context: Callable
     This test reproduces the bug where Arrow format preserves UTC timezone
     in timestamp columns instead of returning naive datetimes.
     """
-    with table_context('test_polars_utc_tz', ['ts DateTime']) as ctx:
+    with table_context('test_polars_utc_tz', ['ts DateTime64']) as ctx:
         test_client.command(f"INSERT INTO {ctx.table} VALUES (now())")
         df = test_client.query_df_arrow(
             f"SELECT * FROM {ctx.table}",


clickhouse_connect/driver/tzutil.py
```diff
@@ -33,6 +33,20 @@ def normalize_timezone(timezone: pytz.timezone) -> Tuple[pytz.timezone, bool]:
     return timezone, False
 
 
+def is_utc_timezone(tz: tzinfo | None) -> bool:
```
joe-clickhouse: I've just started a soft deprecation of Python 3.9 but don't want to immediately break it. The vertical bar syntax for type unions is only available in 3.10+ so if we could please change this to using `typing.Optional`?

First add the `Optional` import from typing and then:

```python
def is_utc_timezone(tz: Optional[tzinfo]) -> bool:
```

clickhouse_connect/driver/client.py
```diff
@@ -30,6 +31,25 @@
 arrow_str_setting = 'output_format_arrow_string_as_string'
 
 
+def _strip_utc_timezone_from_arrow(table: pa.Table) -> pa.Table:
```
joe-clickhouse: Unquoted type hints will also cause a crash if the lib isn't installed so we'll nee to do:

```python
def _strip_utc_timezone_from_arrow(table: "arrow.Table") -> "arrow.Table":
```

joe-clickhouse (changes requested):
Hi @Milias thanks so much for the rebase and update! I've left a few comments that we'll need to address first, but otherwise really good stuff.

clickhouse_connect/driver/tzutil.py
Milias: Oh damn! Yes of course! Too used to Python 3.13 😆

clickhouse_connect/driver/client.py
```diff
@@ -30,6 +31,25 @@
...
+    """Strip UTC timezone from timestamp columns in Arrow table.
+
+    This ensures naive datetimes are returned when the server timezone is UTC
+    and utc_tz_aware is False (the default).
+    """
+    new_fields = []
+    needs_cast = False
+    for field in table.schema:
+        if pa.types.is_timestamp(field.type) and field.type.tz in ('UTC', 'Etc/UTC'):
```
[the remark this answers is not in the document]
Milias: Added now!

Milias:
Sorry, fixed! Should've run it before. I think now it should be fine.

joe-clickhouse (approved):
Looks like we're there! Thanks for the contribution.

PR #639, opened 2026-02-06
Joe/638 fix issue where sending redundant settings is skipped

### LINKED ISSUE #638: Fix issue where sending redundant settings is skipped
Currently, `clickhouse_connect` will silently drop sending settings that already match the server's default. Most of the time this is a sensible thing to do. However, there are some settings such as `do_not_merge_across_partitions_select_final` that according to the [docs](https://clickhouse.com/docs/operations/settings/settings#do_not_merge_across_partitions_select_final)

> If not explicitly set, ClickHouse will automatically enable this optimization when the partition key expression is deterministic and all columns used in the partition key expression are included in the primary key.
>
> Default value: false (but may be automatically enabled based on table structure if not set explicitly)

can be dynamically changed by the server if not explicitly set. Therefore, it's currently impossible to explicitly set the value to 0, the server's default, to prevent the server from changing it.

joe-clickhouse:
## Summary
`clickhouse_connect` had an optimization that skipped sending settings when their values matched the server's default. This is generally fine however, this broke settings where explicitly setting the default has different semantics than leaving it unset.

E.g. for some settings like `do_not_merge_across_partitions_select_final`, ClickHouse distinguishes between:
- Implicitly defaulted to 0 (not explicitly set): In this case, according to [the docs](https://clickhouse.com/docs/operations/settings/settings#do_not_merge_across_partitions_select_final), ClickHouse may auto-enable the optimization based on table structure
- Explicitly set to 0: ClickHouse respects the user's choice and does not auto-enable

Therefore the client incorrectly treated these as equivalent:

```python
if server_default == 0 and user_wants == 0:
    # "It's already 0, no need to send"
```

This PR modifies the validation logic so that settings matching the server default are only skipped when:
1. The setting is readonly and therefore the value can't be changed anyway, so sending it is pointless. This also prevents errors in environments like ClickHouse Cloud where certain settings may be writable locally but readonly in cloud

OR

2. The setting is already explicitly stored on the client to that value

Closes #638

## Checklist


genzgd:
Maybe we could look at the "changed" column in the system.settings table instead of ignoring the server settings entirely?  The query string with a bunch of settings is already bloated.

genzgd:
Whitelisting is a little scary, although hopefully these weird quantum settings are very rare.  (It seems like they could have implemented this with 'yes' 'no' 'dynamic' instead). I thought changed might be set if the setting was non-default for any reason (like profiles) but that is probably wishful thinking.

Issue #567, opened 2025-10-03
Typing errors when running mypy on code that uses clickhouse-connect


chriswithers-fuse:
### Steps to reproduce

```
$ uv run mypy --version
mypy 1.15.0 (compiled: yes)
```

Any code which imports Client will complain with:
```
$ uv run mypy .
...: error: Module "clickhouse_connect.driver" does not explicitly export attribute "Client"  [attr-defined]
```

### Expected behaviour

No mypy complaint

### Configuration

```
$ uv run mypy --version
mypy 1.15.0 (compiled: yes)
$ uv tree | grep connect
Resolved 200 packages in 5ms
├── clickhouse-connect v0.9.2
$ uv run python --version
Python 3.12.11
```

joe-clickhouse:
Hi @chriswithers-fuse thanks! Yeah, taking a look I think `py.typed` was added with good intention in https://github.com/ClickHouse/clickhouse-connect/pull/243 and appears to have worked for the specific use case tested at the time. We currently don't run any mypy testing in CI (once we officially support mypy type checking this will need to change) and it looks like the original validation didn't cover the full public API surface. I'll have to take a survey of what it'd take to get everything fully and correctly type-hinted or decide to just remove that file in the meantime.

PR #643, opened 2026-02-08
Fix mypy typing errors by explicitly exporting Client

veeceey:
## Summary
- Fixes a specific mypy complain by adding `__all__` to `clickhouse_connect.driver` module
- Explicitly exports `Client`, `AsyncClient`, `create_client`, and `create_async_client`

## Problem
Previously, importing `Client` from `clickhouse_connect.driver` would fail mypy validation with:
```
error: Module "clickhouse_connect.driver" does not explicitly export attribute "Client"  [attr-defined]
```

This occurred because the module didn't define `__all__`, so mypy couldn't determine which symbols were intended to be part of the public API.

## Solution
Added `__all__` declaration to explicitly list the public API exports. This makes it clear to type checkers which symbols are intended for external use.

## Test Plan
- Created test file that imports `Client` from the module
- Verified mypy no longer reports the attr-defined error

The fix follows Python best practices for defining module public APIs and is the standard solution for this type of mypy issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,17 +30,20 @@ A `DeprecationWarning` will now be displayed when initializing the client on Pyt
 Python 3.10+ as 3.9 compatibility may break unexpectedly in future updates.
 
 ### Bug Fixes
+- Raise OperationalError when ResponseSource hits network failure before any data is received. Previously, empty result would be returned. Closes [#620](https://github.com/ClickHouse/clickhouse-connect/issues/620)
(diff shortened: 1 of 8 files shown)


joe-clickhouse (approved):
This looks fine. Note that our full public API is not fully typed yet so we're not claiming or committing to full mypy compliance, but this does help resolve one specific mypy complaint. Thanks!

Note that I removed the "Fixes" reference from your description since I want to leave #567 open as this PR address some of the typing issues but not entirely.

PR #644, opened 2026-02-08
Fix percent sign double encoding in SQLAlchemy string literals

### LINKED ISSUE #297: Percent in string literal incorrectly double encoded
### Describe the bug

Percent in string literal incorrectly double encoded with sqlachemy interface.

### Steps to reproduce
1. Make a sql call via sqlalchemy query with a 'Format' in a literal string
e.g. 
`SELECT formatDateTime(toDate('2010-01-04'), '%g')
`

Result:
```

┌─formatDateTime(toDate('2010-01-04'), '%%g')─┐
│ %g                                          │
└─────────────────────────────────────────────┘

```

### Expected behaviour
```
┌─formatDateTime(toDate('2010-01-04'), '%g')─┐
│ 10                                         │
└────────────────────────────────────────────┘
```

### Configuration
On a slightly older version, don't see the necessary fix on master.

### Analysis 

Various places in code has:

```
        if self.preparer._double_percents:
            text = text.replace("%", "%%")
        return text
```

class ChIdentifierPreparer(IdentifierPreparer)

doesn't override the assignment in its base class:

```
        self._double_percents = self.dialect.paramstyle in (
            "format",
            "pyformat",
        )
```

### Workaround
I monkey patched:

```
original_init = ChIdentifierPreparer.__init__

def new_init(self, *args, **kwargs):
    original_init(self, *args, **kwargs)
    self._double_percents = False

ChIdentifierPreparer.__init__ = new_init
```

veeceey:
## Summary
- Fixes #297 by disabling percent doubling in `ChIdentifierPreparer`
- Resolves issue where `%` in SQL string literals were incorrectly escaped to `%%`

## Problem
When using the SQLAlchemy interface, percent signs in SQL string literals were being incorrectly doubled. For example:

```python
SELECT formatDateTime(toDate('2010-01-04'), '%g')
```

Was being transformed to:

```sql
SELECT formatDateTime(toDate('2010-01-04'), '%%g')
```

This broke functions like `formatDateTime` that use `%` for format strings.

## Root Cause
`ChIdentifierPreparer` inherits from SQLAlchemy's `IdentifierPreparer`, which sets `_double_percents = True` for 'format' and 'pyformat' paramstyles. However, ClickHouse doesn't use `%` for parameter placeholders, so this doubling is inappropriate and breaks string literals containing `%`.

## Solution
Override `__init__` to explicitly set `_double_percents = False`, preventing unwanted percent escaping while maintaining all other inherited behavior.

## Test Plan
- Manually verified that queries with `%` in string literals (like `formatDateTime`) now work correctly
- The fix is minimal and only affects percent handling in the SQLAlchemy preparer
- Existing tests continue to pass


joe-clickhouse:
Hi @veeceey apologies for the delay! I'll be able to review this this week. I'm sure you'll see notifications here as I get to it. Thanks again!

veeceey:
@joe-clickhouse No worries at all, thanks for letting me know! Looking forward to your review whenever you get to it.

veeceey:
Hey, added the CHANGELOG entry and fixed the pylint W0212 warning with an inline disable. Thanks for the review!

Issue #534, opened 2025-08-08
Inserting to variant containing String will lead to only String being inserted


SkytAsul:
### Describe the bug
If I correctly understood, currently ClickHouse-connect passes Variant data as String so ClickHouse server infers the type by itself. However, if the Variant contains the String type (e.g. `Variant(Int64, Float32, String)`), all data will be inserted as String.

### Steps to reproduce
1. Create a table containing a `Variant(Int64, Float32, String)` column
2. Insert some int or float data using ClickHouse-connect
3. Run `SELECT variantType(yourColumn)` and see that all rows have String type

### Expected behaviour
The correct type is kept

#### Environment
* clickhouse-connect version: 0.8.18
* Python version: 3.12.3

#### ClickHouse server
* ClickHouse Server version: 25.7.1
* ClickHouse Server non-default settings, if any: 
```csv
"profile","max_threads","1"
"profile","log_queries","1"
"profile","log_queries_min_type","QUERY_FINISH"
"profile","log_queries_min_query_duration_ms","0"
"profile","log_queries_cut_to_length","10000"
"profile","http_max_field_value_size","262144"
"profile","allow_introspection_functions","1"
"profile","use_query_cache","0"
"profile","query_cache_nondeterministic_function_handling","ignore"
"profile","query_cache_system_table_handling","ignore"
"profile","query_cache_min_query_runs","2"
"profile","query_cache_min_query_duration","100"
"profile","query_cache_ttl","30"
"profile","use_query_condition_cache","0"
"server","max_thread_pool_size","10000"
"server","default_database","default"
"server","max_server_memory_usage","30281777971"
"server","max_server_memory_usage_to_ram_ratio","0.9"
"server","max_concurrent_queries","1000"
"server","uncompressed_cache_size","8589934592"
"server","query_condition_cache_size","106700800"
"server","concurrent_threads_soft_limit_num","0"
"server","concurrent_threads_soft_limit_ratio_to_cores","2"
"server","concurrent_threads_scheduler","fair_round_robin"
"server","async_load_databases","1"
"server","keep_alive_timeout","10"
"server","total_memory_profiler_step","4194304"
"server","total_memory_tracker_sample_probability","0"
"server","validate_tcp_client_information","0"
"server","global_profiler_real_time_period_ns","1000000000"
"server","global_profiler_cpu_time_period_ns","100000000"
"server","dictionaries_lazy_load","1"
"server","wait_dictionaries_load_at_startup","1"
```

[...]

PR #651, opened 2026-02-15
Add native write path for Variant data type

### LINKED ISSUE #514: client.insert fails for column of type Variant(Tuple)
### Steps to reproduce
1.
2.
3.

### Code example
```import clickhouse_connect 
import os
from datetime import datetime

    host=os.environ['CLICKHOUSE_HOST'],
    user='default',
    password=os.environ['CLICKHOUSE_PASSWORD'],
    secure=True,
)

query = """
CREATE TABLE test_variant_table
(
    icao String,
    time DateTime64(3),
    altitude Variant(Int32, String),
    aircraft 
    Variant(Tuple(
        a Int32,
        b Int32,
    )),
) ENGINE = ReplacingMergeTree()
PARTITION BY toYYYYMM(time)
ORDER BY (icao, time)
SETTINGS 
index_granularity = 8192;
"""
client.command("SET enable_variant_type = 1;")
client.command(query)

column_names = ['icao','time','altitude','aircraft']
row = (
    "ABCDEFG",
    datetime(2025, 6, 29, 16, 0, 0, 123000),
    30,
    (310,320)
)
client.insert(
    table='test_variant_table',
    data=[row],
    column_names=column_names,
    
)
```

```
---------------------------------------------------------------------------
DatabaseError                             Traceback (most recent call last)
Cell In[16], line 39
     32 column_names = ['icao','time','altitude','aircraft']
     33 row = (
     34     "ABCDEFG",
     35     datetime(2025, 6, 29, 16, 0, 0, 123000),
     36     30,
     37     (310,320)
     38 )
---> 39 client.insert(
     40     table='test_variant_table',
     41     data=[row],
     42     column_names=column_names,
     43     
     44 )

File ~(redacted)/lib/python3.12/site-packages/clickhouse_connect/driver/client.py:676, in Client.insert(self, table, data, column_names, database, column_types, column_type_names, column_oriented, settings, context, transport_settings)
    674         raise ProgrammingError('Attempting to insert new data with non-empty insert context') from None
    675     context.data = data
--> 676 return self.data_insert(context)

File ~(redacted)/lib/python3.12/site-packages/clickhouse_connect/driver/httpclient.py:284, in HttpClient.data_insert(self, context)
    282 params.update(self._validate_settings(context.settings))
    283 headers = dict_copy(headers, context.transport_settings)
...
--> 402 raise OperationalError(err_str) if retried else DatabaseError(err_str) from None

DatabaseError: HTTPDriver for (redacted) received ClickHouse error code 70
 Code: 70. DB::Exception: Cannot convert type String to Variant(Tuple(a Int32, b Int32)). Conversion to Variant allowed only for types from this Variant: while converting column "aircraft" from type String to type Variant(Tuple(a Int32, b Int32)): While executing Native. (CANNOT_CONVERT_TYPE) (version 25.4.1.37390 (official build))
```
#### Environment
* clickhouse-connect version: 0.8.18
* Python version: 3.12.3
* Operating system: Ubuntu 24.04

#### ClickHouse server
* ClickHouse Server version: 25.4.1.37390 

maciek-rr:
## Summary
Taking a stab at implementing this enhancement https://github.com/ClickHouse/clickhouse-connect/issues/534.

Implement native write path for Variant data type, replacing the previous approach that serialized all values as strings. 

Values are now written in native binary format with automatic type dispatch based on Python types. For ambiguous cases (e.g., `Variant(Array(UInt32), Array(String))` where both map to list), a `typed_variant()` helper allows explicit type tagging. 

When automatic dispatch is ambiguous and no explicit type tag is provided, an exception is raised. This is a backwards-incompatible change - previously the value would have been stored as a string if the Variant contained a String type.                                                                                                                                                                              

## Migration note

Previously, values inserted into `Variant` columns were stringified and sent to the server,
which stored them in the `String` member when available. With this change, values are
dispatched to their native ClickHouse types client-side.

**Before:**
```python
# Variant(Int64, String)
client.insert('t', [[100], ['hello']], column_names=['v'])
# SELECT variantType(v) → 'String', 'String'
```
**After:**
```python
# Variant(Int64, String)
client.insert('t', [[100], ['hello']], column_names=['v'])
# SELECT variantType(v) → 'Int64', 'String'
```

If your variant column contains members that share a Python type (e.g. `Array(UInt32)` and `Array(String)` are both list), use the new typed_variant helper to disambiguate:
```python
from clickhouse_connect.datatypes.dynamic import typed_variant

client.insert('t', [[typed_variant([1, 2], 'Array(UInt32)')],
                    [typed_variant(['a', 'b'], 'Array(String)')]], column_names=['v'])
```

Values that cannot be mapped to any variant member now raise DataError immediately,
rather than being stringified and sent to the server.

Closes #514

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,16 @@ The supported method of passing ClickHouse server settings is to prefix such arg
 ## UNRELEASED
 
 ### Improvements
+- BREAKING CHANGE: Implement native write path for `Variant` data type with type-aware dispatching.
+Previously, all values inserted into a `Variant` column were stringified and sent to the server, which
+would store them in the `String` member if present, or attempt server-side conversion otherwise. Values
+are now serialized using their native ClickHouse types client-side (e.g. inserting `100` into
+`Variant(Int64, String)` stores `Int64(100)` instead of `String("100")`). Key changes:
+  - Values that don't match any variant member now raise `DataError` instead of being stringified and
+  delegated to the server.
+  - A `typed_variant(value, 'TypeName')` helper is provided for cases where automatic dispatch
+  cannot resolve the target type, such as when multiple variant members map to the same Python
+  type (e.g. `Array(UInt32)` vs `Array(String)`).
 - Added `utc_tz_aware="schema"` mode which returns timezone-aware datetimes only when the server's column schema explicitly defines a timezone (e.g. `DateTime('UTC')`), and naive datetimes for bare `DateTime` columns. This matches the ClickHouse schema definition exactly. Not yet supported for Arrow-based query methods. Closes [#645](https://github.com/ClickHouse/clickhouse-connect/issues/645)
 
 ### Bug Fixes
--- a/clickhouse_connect/datatypes/dynamic.py
+++ b/clickhouse_connect/datatypes/dynamic.py
@@ -1,5 +1,5 @@
 from collections import namedtuple
-from typing import List, Sequence, Collection, Any
+from typing import List, Tuple, Sequence, Collection, Any
 from urllib.parse import unquote
 
 from clickhouse_connect.datatypes.base import ClickHouseType, TypeDef
@@ -7,7 +7,7 @@ from clickhouse_connect.datatypes.registry import get_from_name
 from clickhouse_connect.driver.common import unescape_identifier, first_value, write_uint64
 from clickhouse_connect.driver.ctypes import data_conv
 from clickhouse_connect.driver.errors import handle_error
-from clickhouse_connect.driver.exceptions import DataError
+from clickhouse_connect.driver.exceptions import DataError, InternalError
 from clickhouse_connect.driver.insert import InsertContext
 from clickhouse_connect.driver.query import QueryContext
 from clickhouse_connect.driver.types import ByteSource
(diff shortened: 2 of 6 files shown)


joe-clickhouse (commented):
Really good stuff here @maciek-rr! Thanks for doing this. I think it's the right approach, with a solid implementation, and has good test coverage. There are only a few small things I'd like to see:
1. A more significant note in the CHANGELOG. This is a pretty significant breaking change. It's a good breaking change, but previously it was extremely forgiving in that you could basically insert anything into any column and it'd just convert it to a string and hope there was a `String` defined in the `Variant`. Definitely put **BREAKING CHANGE:** at the start of the line and then maybe a bit of an explanation of what/how it's different now. e.g.:
    - Values that don't match a variant member now raise `DataError` instead of being silently stringified
    - Data stored via the old path was all `String`. The same inserts now store native types
    - `typed_variant()` is required when multiple variant members share a Python type
    - Anything else you think the user needs to know about
2. `typed_variant` is part of the public API now, so a docstring would be helpful for IDE autocomplete, etc.
3. If you could add one more section to the PR description like `## Migration note` and then show a concrete before and after. Something like:

    If you were inserting `100` into `Variant(Int64, String)`, the old code stored it as `String("100")`. It now correctly stores it as `Int64(100)`. This means `SELECT variantType(col)` will return `'Int64'` where it previously returned `'String'`. In other words, now if you insert an `int` you'll get an `int` back...which is awesome.

Let me know if you have any questions. Thanks again!

PR #655, opened 2026-02-21
Add GROUP BY alias rendering to ClickHouse SQLAlchemy compiler

joe-clickhouse:
## Summary
ClickHouse resolves `SELECT` aliases inside `GROUP BY` expressions which causes circular reference errors when an alias shadows a source column name. For example

```sql
SELECT toStartOfDay(toDateTime(time)) AS time
...
GROUP BY toStartOfDay(toDateTime(time))
```

fails because `time` in the expression resolves to the alias, not the column.

This PR adds `group_by_clause` and `visit_label` overrides to `ChStatementCompiler` so that labeled expressions in `GROUP BY` render as their alias name instead of the full expression. Note that `SELECT` and `ORDER BY` rendering is unaffected.

This fixes the root cause and underlying issue behind https://github.com/apache/superset/issues/33551

Consequently, this should eliminate the need for Superset's [alias-mangling](https://github.com/apache/superset/pull/23185#issue-1598213991) workaround that appends hash suffixes to all ClickHouse column aliases to avoid these `GROUP BY` collisions.

To fully close the loop here I'll need to close https://github.com/apache/superset/pull/34091 and then open another PR that no-ops or removes the current [`_mutate_label`](https://github.com/apache/superset/blob/987b6a6f0419c9570f3429c78d2322b45a289fa5/superset/db_engine_specs/clickhouse.py#L522) behavior and then require a version of ciikchouse-connect in superset greater than or equal to whatever version we cut this release in.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ are now serialized using their native ClickHouse types client-side (e.g. inserti
 
 ### Bug Fixes
 - Recognize `UPDATE` as a command so lightweight updates work correctly via `client.query()` and SQLAlchemy.
+- SQLAlchemy: `GROUP BY` now renders label aliases instead of full expressions which avoids circular reference errors when an alias shadows a source column name in ClickHouse.
 
 ## 0.11.0, 2026-02-10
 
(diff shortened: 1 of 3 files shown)


villebro (approved):
LGTM, the Superset community will definitely ❤️  this improvement!

PR #707, opened 2026-04-08
fix: support Fixed/UTC timezone format in DateTime/DateTime64

Bahtya:
## Problem

`clickhouse-connect` fails to deserialize `DateTime(Fixed/UTC±HH:MM:SS)` columns:

```python
result = client.query(
    "SELECT toStartOfInterval(now(), toIntervalHour(1), 'Fixed/UTC+05:30:00') AS ts"
)
# InternalError: Unrecognized ClickHouse type base: DateTime name: DateTime('Fixed/UTC+05:30:00')
```

## Root Cause

`DateTime.__init__` and `DateTime64.__init__` use `pytz.timezone()` to parse timezone names. `pytz` does not recognize ClickHouse's `Fixed/UTC±HH:MM:SS` format, causing the deserialization to fail.

## Solution

Add `_parse_timezone()` helper that:
1. Detects `Fixed/UTC±HH:MM:SS` format via regex
2. Creates a `datetime.timezone` fixed-offset for those
3. Falls back to `pytz.timezone()` for standard names

Both `DateTime` and `DateTime64` updated to use the new helper.

Fixes #702

### DIFF
--- a/clickhouse_connect/cc_sqlalchemy/datatypes/sqltypes.py
+++ b/clickhouse_connect/cc_sqlalchemy/datatypes/sqltypes.py
@@ -91,9 +91,9 @@ class Float64(ChSqlaType, Float):
 
 
 class Bool(ChSqlaType, SqlaBoolean):
-    def __init__(self, type_def: TypeDef = EMPTY_TYPE_DEF):
+    def __init__(self, type_def: TypeDef = EMPTY_TYPE_DEF, **kwargs):
         ChSqlaType.__init__(self, type_def)
-        SqlaBoolean.__init__(self)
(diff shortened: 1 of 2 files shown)


Bahtya:
I have read the CLA Document and I hereby sign the CLA.

joe-clickhouse:
Hi @Bahtya thanks for this! I really appreciate the time and effort, but I'm actually about to rework a portion of the timezone-related stuff in the lib really soon and already have a fix for it on a local branch so it'll get addressed when that effort is merged in. I'll mark #702 as in progress so people know. Thanks again!

PR #709, opened 2026-04-15
fix: support Fixed/UTC±HH:MM:SS timezone format in DateTime and DateTime64

shubham1166:
## Summary

Closes #702

ClickHouse uses a `Fixed` offset timezone format — e.g. `DateTime('Fixed/UTC+05:30:00')` — in environments where the IANA timezone database is unavailable on the server. The Python `pytz` library does not recognise this prefix, causing a hard crash when the client tries to deserialise any `DateTime` or `DateTime64` column that carries a `Fixed` timezone:

```
clickhouse_connect.driver.exceptions.InternalError:
  Unrecognized ClickHouse type base: DateTime name: DateTime('Fixed/UTC+05:30:00')
```

Users cannot query those columns at all — it is a complete failure, not a degraded result.

---

## Root Cause

In `temporal.py`, both `DateTime.__init__` and `DateTime64.__init__` pass the raw timezone string directly to `pytz.timezone()`:

```python
# Before
self.tzinfo = pytz.timezone(type_def.values[0][1:-1])
```

`pytz` does not know about `Fixed/UTC+05:30:00`, raises `UnknownTimeZoneError`, and the registry then re-raises it as an `InternalError`.

---

## Fix

Added a `parse_timezone(tz_str)` helper to `clickhouse_connect/driver/tzutil.py`:

- Detects the `Fixed/UTC±HH:MM:SS` pattern with a compiled regex
- Parses the sign, hours, minutes, and seconds components
- Returns a stdlib `datetime.timezone` built from a `timedelta`
- All other timezone strings fall through to `pytz.timezone()` unchanged — **zero behaviour change for existing code**

Replaced the two direct `pytz.timezone()` calls in `DateTime.__init__` and `DateTime64.__init__` with `tzutil.parse_timezone()`.

**Total change: 15 lines across 2 source files.**

---

### Unit tests — all passing (`tests/unit_tests/test_chtypes.py`)

8 new tests were added covering the full surface of `parse_timezone()` and the two affected type constructors:

| Test | Verifies |
|---|---|
| `test_parse_timezone_fixed_positive` | `Fixed/UTC+05:30:00` → correct `timedelta` |
| `test_parse_timezone_fixed_negative` | `Fixed/UTC-03:30:00` → correct negative `timedelta` |
| `test_parse_timezone_fixed_zero` | `Fixed/UTC+00:00:00` → zero offset |
| `test_parse_timezone_fixed_with_seconds` | `Fixed/UTC+05:30:30` → seconds component parsed |
| `test_parse_timezone_standard_pytz` | `America/Chicago` still resolves via pytz |
| `test_datetime_fixed_timezone` | `DateTime('Fixed/UTC+05:30:00')` instantiates correctly |
| `test_datetime_fixed_timezone_negative` | `DateTime('Fixed/UTC-03:00:00')` instantiates correctly |
| `test_datetime64_fixed_timezone` | `DateTime64(3, 'Fixed/UTC+05:30:00')` instantiates correctly |

Full unit suite result (excluding pre-existing errors unrelated to this change):
```
263 passed, 5 warnings  ✅  (+8 new tests vs 255 on main)
```

### Integration tests — all passing against ClickHouse 26.3.9

All 20 existing timezone integration tests pass with no changes:
```
tests/integration_tests/test_timezones.py  20 passed  ✅
```

### End-to-end proof against a live ClickHouse 26.3.9 instance

The exact query from the issue now works correctly:

```python
# Before fix → InternalError: Unrecognized ClickHouse type base: DateTime
# After fix  → works correctly

client.query("SELECT toDateTime('2024-01-15 10:30:00', 'Fixed/UTC+05:30:00') AS dt")
# Result: 2024-01-15 10:30:00+05:30  ✅

client.query("SELECT toDateTime('2024-01-15 10:30:00', 'Fixed/UTC-03:00:00') AS dt")
# Result: 2024-01-15 10:30:00-03:00  ✅

client.query("SELECT toDateTime64('2024-01-15 10:30:00.123', 3, 'Fixed/UTC+05:30:00') AS dt64")
# Result: 2024-01-15 10:30:00.123000+05:30  ✅
```

---

## Files Changed

| File | Change |
|---|---|
| `clickhouse_connect/driver/tzutil.py` | Added `parse_timezone()` helper + `_FIXED_TZ_RE` regex |
| `clickhouse_connect/datatypes/temporal.py` | Use `tzutil.parse_timezone()` in `DateTime` and `DateTime64` |
| `tests/unit_tests/test_chtypes.py` | 8 new unit tests |


joe-clickhouse:
Hi @shubham1166 thanks for the work so far! There are a few things we still want to tighten up for correctness and completeness here.

## Other call sites
There are four other call sites that also parse a clickhouse-sourced timezone string. Your PR description calls out "environments where the IANA timezone database is unavailable on the server." On those servers, `SELECT timezone()` will return `Fixed/UTC+HH:MM:SS` and the `X-ClickHouse-Timezone` header carries the same format on every response. Consequently the following sites hit the same problem:

1. `clickhouse_connect/driver/client.py:203` in `_init_common_settings`. This is what parses the `SELECT timezone()` during client init. You'll want to replace `pytz.timezone(server_tz)` with `tzutil.parse_timezone(server_tz)`.
2. `clickhouse_connect/driver/asyncclient.py:314` you'll need to do the same thing here for the async client.
3. `clickhouse_connect/driver/client.py:299` in `_check_tz_change` you'll also want to replace `pytz.timezone(new_tz)` with `tzutil.parse_timezone(new_tz)`.
4. `clickhouse_connect/driver/binding.py:72` in `_extract_tz_from_type` you'll also need to replace `pytz.timezone(v[1:-1])` with `tzutil.parse_timezone(v[1:-1])`

## Unit tests
Then, if you can please add at least 2 unit tests exercising the fixed tz path. i.e. test `_extract_tz_from_type` with a string like `"DateTime('Fixed/UTC+05:30:00')"` and then also `_check_tz_change` with something like `"Fixed/UTC+05:30:00"` as a string asserting that we end up with a non-UTC fixed-offset tzinfo.

## Short circuit Fixed offset tzs in `normalize_timezone`
Please add an `isinstance(tz, datetime.timezone)` short-circuit at the top of `tz_util.normalize_timezone()` returning `(tz, True)`.

<details>
<summary>Click here for the long explanation</summary>
Once call site 1 (<code>clickhouse_connect/driver/client.py:203</code>) is handled and <code>_init_common_settings</code> correctly parses the Fixed offset into a stdlib <code>datetime.timezone</code>, it then flows into <code>tzutil.normalize_timezone()</code>. That function checks <code>tz.tzname(None) in pytz.common_timezones</code> to decide <code>_dst_safe</code>, and a stdlib <code>datetime.timezone</code> reports something like <code>"UTC+05:30"</code> which isn't in that set. Consequently, <code>_dst_safe</code> gets set to <code>False</code>, and under the default <code>tz_source="auto"</code> this flips <code>_apply_server_tz=False</code>. So the server's Fixed tz gets correctly parsed but then silently not applied. However, Fixed offsets are by definition DST-safe and self-describing. So we need to tell <code>normalize_timezone()</code> to trust Fixed offsets.
</details>

## Don't modify
Leave the remaining `pytz.timezone()` calls in `query.py:124,132` and `cc_sqlalchemy/datatypes/sqltypes.py:268,286` alone. Those take user-supplied strings where raising on unknown is the correct behavior.

shubham1166:
Thank you for having a look, i will have a look by tomorrow evening and update it, i hope that is fine on you end. Thanks again.

joe-clickhouse:
Thanks @shubham1166! No problem. Take your time and let me know if you need any clarification or have questions or need help with anything.

joe-clickhouse:
Hi @shubham1166! Thanks for you efforts this. Since we started working on this PR, some significant changes have been merged into main. Specifically, we've replaced `pytz` with the stdlib `zoneinfo` lib and consequently, the implementation needed to address #702 has changed significantly so I'm going to close this PR down as the issues is now resolved in https://github.com/ClickHouse/clickhouse-connect/pull/746. Thank you again for the work though! Happy to help support you through any other PR you're willing to work on.

PR #726, opened 2026-04-30
async client: retry on bare ServerDisconnectedError

### LINKED ISSUE #725: Async client raises OperationalError on stale pooled connection instead of retrying
### Describe the bug
The async client (`get_async_client`) raises `OperationalError("Network Error: Server disconnected")` on the first request after a pooled keep-alive connection has been closed by the server or an upstream proxy. The retry path in `_raw_request` (`clickhouse_connect/driver/asyncclient.py`) covers `"Connection reset"`, `"Remote end closed"`, and `"Cannot connect"`, but not the default `"Server disconnected"` message that `aiohttp.ServerDisconnectedError()` raises in current aiohttp versions.

In production behind an HTTP load balancer with idle keep-alive timeouts shorter than the client's pool TTL, this surfaces as a transient error on the first request after any idle window, even though a single retry on a fresh connection would succeed.

The sync (urllib3) path is unaffected because urllib3's `Retry` handles dropped pooled connections internally.

### Steps to reproduce
1. Open an async client and run a query (which warms a pooled connection).
2. Force the next `aiohttp.ClientSession.request` call to raise `aiohttp.ServerDisconnectedError()` (the same class and default message aiohttp raises for stale pooled connections).
3. Issue another query.

### Expected behaviour
The disconnect is treated like the existing `"Connection reset"` / `"Remote end closed"` cases: retried once on a fresh connection, transparent to the caller.

### Code example
```python
import asyncio
from unittest.mock import patch
import aiohttp
import clickhouse_connect

async def main():
    client = await clickhouse_connect.get_async_client(host="localhost", port=8123)
    await client.query("SELECT 1")  # warms the pool

    with patch.object(
        client._session,
        "request",
        side_effect=aiohttp.ServerDisconnectedError(),
    ):
        await client.query("SELECT 1")  # raises OperationalError("Network Error: Server disconnected")

asyncio.run(main())
```

#### Environment
* clickhouse-connect version: 1.0.0rc1 (also reproduces on 0.12.0rc1)
* Python version: 3.12
* Operating system: Ubuntu (production), macOS (local repro)

#### ClickHouse server
* Any version reachable over HTTP; the disconnect is raised by aiohttp on the client side, not by ClickHouse.

---

I am opening a PR to address the issue.

IAL32:
## Summary
The async client's `_raw_request` retry path catches `aiohttp.ServerConnectionError` and retries once when `str(e)` contains `"Connection reset"`, `"Remote end closed"`, or `"Cannot connect"`. It does not match the bare `aiohttp.ServerDisconnectedError()` whose default message is `"Server disconnected"`, which is what aiohttp raises in current versions when a pooled keep-alive connection has been closed by the server (or an upstream load balancer) between requests. The first request after an idle period reuses the stale connection, sees the disconnect, and surfaces as `OperationalError("Network Error: Server disconnected")` to the caller instead of being transparently retried on a fresh connection.

The fix adds `"Server disconnected"` to the existing substring list. Reproduces and verifies in `tests/integration_tests/test_error_handling.py` against a real ClickHouse server using the `test_native_async_client` fixture and `pytest-mock` to inject the disconnect at the `_session.request` boundary.

Tests added:
- `test_async_retry_on_server_disconnected`: bare `ServerDisconnectedError()` is retried once and the second attempt succeeds. Fails on `main`, passes with this change.
- `test_async_server_disconnected_raises_after_retry`: if the disconnect persists, `OperationalError("Network Error: Server disconnected")` still surfaces with the original `ServerDisconnectedError` as `__cause__`.
- `test_async_retry_on_connection_reset`: regression check that the pre-existing `"Connection reset"` retry branch still works.

This is async-only. The sync client uses urllib3, which retries pooled-connection drops independently via `Retry`.

Closes https://github.com/ClickHouse/clickhouse-connect/issues/725

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## UNRELEASED
 
+### Bug Fixes
+- Async client: retry once when a pooled keep-alive connection is closed by the server and aiohttp raises `ServerDisconnectedError` with the default `"Server disconnected"` message. The existing retry path covered `"Connection reset"` and `"Remote end closed"`, but not the bare `ServerDisconnectedError()` produced by recent aiohttp versions, which surfaced as an `OperationalError("Network Error: Server disconnected")` on the first request after an idle period.
+
 ## 1.0.0rc1, 2026-04-22
 
 ### Breaking Changes
--- a/clickhouse_connect/driver/asyncclient.py
+++ b/clickhouse_connect/driver/asyncclient.py
@@ -1857,11 +1857,12 @@ class AsyncClient(Client):
                 await self._error_handler(response)
 
             except aiohttp.ServerConnectionError as e:
-                if "Connection reset" in str(e) or "Remote end closed" in str(e) or "Cannot connect" in str(e):
+                msg = str(e)
+                if "Connection reset" in msg or "Remote end closed" in msg or "Cannot connect" in msg or "Server disconnected" in msg:
                     if attempts == 1:
                         logger.debug("Retrying after connection error from remote host")
                         continue
-                raise OperationalError(f"Network Error: {str(e)}") from e
+                raise OperationalError(f"Network Error: {msg}") from e
 
             except (aiohttp.ClientError, asyncio.TimeoutError) as e:
                 raise OperationalError(f"Network Error: {str(e)}") from e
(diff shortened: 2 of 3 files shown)


joe-clickhouse (approved):
Looks good. Thanks!

PR #727, opened 2026-05-04
Use SPDX license identifier

Shortfinga:
The SPDX ID for the Apache 2 is Apache-2.0. This would help compliance and license scanners to determine the correct license.

I'm not sure if this deserves mentioning in the Changelog, I think not.


joe-clickhouse (approved):
@Shortfinga Looks good! I need to do some work to be fully [PEP 639](https://peps.python.org/pep-0639/) compliant but I'll get there eventually. Either way, switching to the short identifier is good for now. If you could just sign the CLA when you have a moment, I can get this merged. Thanks!

PR #734, opened 2026-05-07
Fix SQLAlchemy ILIKE and NOT ILIKE rendering

bialkou:
## Summary
Fixes SQLAlchemy dialect rendering for `ILIKE` and `NOT ILIKE` expressions.

SQLAlchemy's generic compiler fallback renders `ilike()` / `not_ilike()` as `lower(...) LIKE lower(...)` / `lower(...) NOT LIKE lower(...)`, which can break matching logic for non-ASCII text. ClickHouse supports native `ILIKE`, so the dialect now renders these expressions directly as `ILIKE` and `NOT ILIKE`.

## Checklist
- [x] Unit tests covering the common scenarios were added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ### Bug Fixes
 - Fix intermittent `Code: 62. Empty query. (SYNTAX_ERROR)` on inserts when a pooled keep-alive connection is reset between attempts. The retry path now rebuilds the insert body instead of replaying an already-drained generator. Affects both sync and async clients. Closes [#731](https://github.com/ClickHouse/clickhouse-connect/issues/731)
+- Fix SQLAlchemy dialect rendering for `ILIKE` and `NOT ILIKE` expressions to use native ClickHouse syntax instead of the generic SQLAlchemy `lower(...) LIKE lower(...)` fallback.
 
 ## 1.0.0rc2, 2026-05-05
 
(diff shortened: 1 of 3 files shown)


joe-clickhouse (changes requested):
Hi @bialkou thanks for the PR! I think this is the right way to do this, however, there's one subtlety we need to consider. [SQLAlchemy's](https://docs.sqlalchemy.org/en/21/core/sqlelement.html#sqlalchemy.sql.expression.ColumnElement.ilike) `like()` and `ilike()` can take an `escape=` modifier. However, ClickHouse server does not have an `ESCAPE` clause on `LIKE` or `ILIKE`. Currently, on main you can supply escape but the server will return a syntax error. In this implementation, escape is silently ignored. In other words, the behavior changes.

The correct behavior now should be a compile error since we can catch it client-side. So I think before the returns in the `like` and `ilike` function, add a guard or something like

```python
if binary.modifiers.get("escape") is not None:
    raise CompileError("ClickHouse does not support the ESCAPE clause on ILIKE")
```

and one for `LIKE` as well. This way, the error behavior is consistent (and correct) with what currently happens. Then if you could add a quick unit test showing this happens, that'd be great. Thanks!

bialkou:
@joe-clickhouse thanks, fixed it!
Added compile-time validation for `escape=` on `LIKE`, `NOT LIKE`, `ILIKE`, and `NOT ILIKE` with `CompileError` and  unit test for the unsupported `escape=` cases.

joe-clickhouse:
Thanks @bialkou! I'm in the process of releasing 1.0.0 today so I'll get this wrapped up into the first bugfix release.

joe-clickhouse (approved):

joe-clickhouse:
@bialkou [v1.0.1](https://github.com/ClickHouse/clickhouse-connect/releases/tag/v1.0.1) has been released with this fix

PR #755, opened 2026-05-21
async client: retry ServerDisconnectedErrors up to retries

### LINKED ISSUE #754: Async client raises OperationalError on consecutive ServerDisconnectedErrors instead of retrying
### Describe the bug
My PR #726 added `"Server disconnected"` to the retry substring list in `_raw_request` (`clickhouse_connect/driver/asyncclient.py`), but the retry is gated on `attempts == 1`. The async client retries exactly once on a network error regardless of `query_retries`, while the 429/503/504 branch in the same function drains the full `retries` budget.

When two consecutive `aiohttp.ClientSession.request` calls raise `aiohttp.ServerDisconnectedError`, the second one surfaces as `OperationalError("Network Error: Server disconnected")` to the caller instead of being retried.

The sync (urllib3) path is unaffected.

  ### Steps to reproduce
  1. Open an async client and warm the pool.
  2. Force the next two `aiohttp.ClientSession.request` calls to raise `aiohttp.ServerDisconnectedError()`.
  3. Issue a query.

  ### Expected behaviour
  Network-error retries drain the same `retries` budget as 429/503/504 retries. With `query_retries=2`, a third attempt on a fresh connection succeeds, transparent to the caller.

  ### Code example
  ```python
import asyncio
import clickhouse_connect

drops = [0]


async def proxy(src_reader, src_writer):
    if drops[0]:
        drops[0] -= 1
        await src_reader.read(4096)
        src_writer.close()
        return
    dst_reader, dst_writer = await asyncio.open_connection("localhost", 8123)

    async def pipe(r, w):
        try:
            while data := await r.read(4096):
                w.write(data)
                await w.drain()
        except OSError:
            pass

    await asyncio.gather(pipe(src_reader, dst_writer), pipe(dst_reader, src_writer))
    src_writer.close()
    dst_writer.close()


async def main():
    server = await asyncio.start_server(proxy, "127.0.0.1", 18123)
    asyncio.create_task(server.serve_forever())
    async with await clickhouse_connect.get_async_client(
        host="127.0.0.1", port=18123, keepalive_timeout=0
    ) as client:
        drops[0] = 2
        await client.query("SELECT 1")  # raises OperationalError on 1.0.1


asyncio.run(main())
  ```

Here we spin up a TCP proxy that drops the first two client connections (graceful close after reading the request, before any response) and forwards the rest to a real ClickHouse, then issues a query through it: on 1.0.1 the second drop exhausts the one-retry budget and the query raises OperationalError.

  ### Configuration
  #### Environment
  * clickhouse-connect version: 1.0.1
  * Python version: 3.12
  * Operating system: Ubuntu (production), macOS (local repro)

  #### ClickHouse server
  * Any version reachable over HTTP; the disconnect is raised by aiohttp on the client side.

  ---

  I am opening a PR to address the issue.

IAL32:
PR #726 added "Server disconnected" to the retry substring list in `_raw_request` but kept the existing attempts == 1 gate, so the async client retries at most once on a network error regardless of
  query_retries. Two consecutive ServerDisconnectedErrors still surface as OperationalError to the caller. The 429/503/504 branch above already drains the retries budget; this lines up the two.

  The fix drains the same retries budget on network errors as on HTTP-status retries, while preserving the existing one-retry guarantee for callers that pass `retries=0` (the insert path).

  `test_async_retry_on_server_disconnected` is parametrized over disconnect count and message variant. The single-disconnect cases preserve the regression for #726; the two-disconnect cases fail on main and
  pass with this change. The previously separate test_async_retry_on_connection_reset is folded into the parametrization.

  Async-only. Sync (`urllib3`) retries pooled-connection drops independently via Retry.

  Closes #754 

  ## Checklist


joe-clickhouse (commented):
@IAL32 thanks! Looks good. Three small comments:
1. Please add a CHANGELOG entry
2. Might be worth adding a small i.e. 0.1s backoff between connection-error retries? This is was the status code branch does, so let's make it consistent.
3. Please extend this to mirror the fix in the sync client as well. It'd be around https://github.com/ClickHouse/clickhouse-connect/blob/2ad7ef33bcb50387a91a134486bbc8760355ce52/clickhouse_connect/driver/httpclient.py#L571

IAL32:
@joe-clickhouse all fair points! I updated the PR. Thanks!

joe-clickhouse:
Thanks! Looks good now. Hope you don't mind, but I pushed a small change on top. Sync `raw_query` and `raw_stream` weren't passing `retries=self.query_retries` to `_raw_request`, so they were still effectively capped at one retry on stale-connection bursts. This was a pre-existing parity gap but worth wrapping into this. Also added one test.

Again, thanks for the contribution!

PR #766, opened 2026-05-27
Implement SQLAlchemy reflection for the cc_sqlalchemy dialect

coby-astsec:
## Summary

`cc_sqlalchemy` doesn't implement SQLAlchemy reflection — `MetaData.reflect()` and `Inspector.get_multi_columns()` raise `NotImplementedError` on any CH database. This breaks tools like `sqlacodegen` or other tools which introspect schema.

Mechanism: SQLAlchemy's reflection path is `Inspector -> Dialect.get_multi_columns -> Dialect.get_columns`. The dialect only ever defined `Inspector.get_columns` (on `ChInspector`), which `MetaData.reflect()` never calls on its own. Reflection falls through to the `DefaultDialect` base, which raises.

This PR fills in the missing dialect method. No runtime client paths change.

### Changes

1. **`dialect.py`** — add `ClickHouseDialect.get_columns()`.

2. **`inspector.py`** — promote `get_columns` to a module-level function shared between dialect and inspector.

3. **`datatypes/sqltypes.py` — concrete `python_type`s.** SQLAlchemy's `TypeEngine.python_type` contract is "return a class or raise `NotImplementedError`." Returning `None` (the current behavior on every UDT-based type) makes `python_type.__module__` / `.__name__` raise `AttributeError`, which is what breaks sqlacodegen:

   | Type | `python_type` |
   |---|---|
   | `UUID` | `uuid.UUID` |
   | `IPv4` / `IPv6` | `ipaddress.IPv4Address` / `IPv6Address` |
   | `Nothing` | `type(None)` |
   | `Point` | `tuple` |
   | `Ring` / `Polygon` / `MultiPolygon` / `LineString` / `MultiLineString` | `list` |
   | `JSON` | `dict` |
   | `Nested` | `list` |
   | `(Simple)AggregateFunction` | `str` |

4. **`datatypes/sqltypes.py` — `Array` now subclasses `sqlalchemy.types.ARRAY`** alongside `ChSqlaType`, exposes `item_type` as a plain instance attribute (so sqlacodegen's `fix_column_types` adaptation pass can reassign it), and sets `dimensions = 1`. `ARRAY.__init__` is *not* called cooperatively because it rejects nested ARRAY `item_type`s, which CH supports natively (`Array(Array(T))`).

### Tests

`tests/integration_tests/test_sqlalchemy/test_reflect.py` covers the dialect path (`MetaData.reflect()`) and a direct `Table(autoload_with=...)` call against a MergeTree with `ORDER BY (org_id, id)`, plus a user-declared composite primary key surviving reflection.

Local: SQLAlchemy tests pass.

### Checklist

- [x] CHANGELOG entry included
- [x] Docs - n/a

### DIFF
--- a/clickhouse_connect/cc_sqlalchemy/datatypes/sqltypes.py
+++ b/clickhouse_connect/cc_sqlalchemy/datatypes/sqltypes.py
@@ -1,7 +1,17 @@
+import ipaddress
+import uuid
 from collections.abc import Sequence
 from enum import Enum as PyEnum
 
 from sqlalchemy.exc import ArgumentError
+from sqlalchemy.types import (
+    ARRAY,
+    Float,
+    Integer,
+    Interval,
+    Numeric,
+    UserDefinedType,
+)
 from sqlalchemy.types import (
     Boolean as SqlaBoolean,
 )
@@ -12,17 +22,10 @@ from sqlalchemy.types import (
     DateTime as SqlaDateTime,
 )
 from sqlalchemy.types import (
-    Float,
-    Integer,
-    Interval,
-    Numeric,
-    UserDefinedType,
-)
-from sqlalchemy.types import (
     String as SqlaString,
 )
 
-from clickhouse_connect.cc_sqlalchemy.datatypes.base import ChSqlaType, schema_types
+from clickhouse_connect.cc_sqlalchemy.datatypes.base import ChSqlaType, schema_types, sqla_type_from_name
 from clickhouse_connect.datatypes.base import EMPTY_TYPE_DEF, LC_TYPE_DEF, NULLABLE_TYPE_DEF, TypeDef
 from clickhouse_connect.datatypes.numeric import Enum8 as ChEnum8
 from clickhouse_connect.datatypes.numeric import Enum16 as ChEnum16
@@ -209,43 +212,43 @@ class FixedString(ChSqlaType, SqlaString):
 
 
 class IPv4(ChSqlaType, UserDefinedType):
-    python_type = None
+    python_type = ipaddress.IPv4Address
 
 
 class IPv6(ChSqlaType, UserDefinedType):
-    python_type = None
+    python_type = ipaddress.IPv6Address
 
 
 class UUID(ChSqlaType, UserDefinedType):
-    python_type = None
+    python_type = uuid.UUID
 
 
 class Nothing(ChSqlaType, UserDefinedType):
-    python_type = None
+    python_type = type(None)
 
 
 class Point(ChSqlaType, UserDefinedType):
-    python_type = None
+    python_type = tuple
 
 
 class Ring(ChSqlaType, UserDefinedType):
-    python_type = None
+    python_type = list
 
 
 class Polygon(ChSqlaType, UserDefinedType):
-    python_type = None
+    python_type = list
 
 
 class MultiPolygon(ChSqlaType, UserDefinedType):
-    python_type = None
+    python_type = list
 
 
 class LineString(ChSqlaType, UserDefinedType):
-    python_type = None
+    python_type = list
 
 
 class MultiLineString(ChSqlaType, UserDefinedType):
-    python_type = None
+    python_type = list
 
 
 class Date(ChSqlaType, SqlaDate):
@@ -412,8 +415,9 @@ class LowCardinality:
         return element.__class__(type_def=TypeDef(wrappers, orig.keys, orig.values))
 
 
-class Array(ChSqlaType, UserDefinedType):
+class Array(ChSqlaType, ARRAY):
     python_type = list
+    dimensions = 1
 
     def __init__(self, element: ChSqlaType | type[ChSqlaType] = None, type_def: TypeDef = None):
         """
@@ -425,7 +429,10 @@ class Array(ChSqlaType, UserDefinedType):
             if callable(element):
                 element = element()
             type_def = TypeDef(values=(element.name,))
-        super().__init__(type_def)
+        ChSqlaType.__init__(self, type_def)
+        # Set item_type directly; calling ARRAY.__init__ would reject nested Array(Array(T)),
+        # which CH supports natively (CH expresses dimensions via nesting, not a dim count).
+        self.item_type = sqla_type_from_name(type_def.values[0])
 
 
 class Map(ChSqlaType, UserDefinedType):
@@ -489,7 +496,7 @@ class JSON(ChSqlaType, UserDefinedType):
     Note this isn't currently supported for insert/select, only table definitions
     """
 
-    python_type = None
+    python_type = dict
 
 
 class Nested(ChSqlaType, UserDefinedType):
@@ -497,11 +504,11 @@ class Nested(ChSqlaType, UserDefinedType):
     Note this isn't currently supported for insert/select, only table definitions
     """
 
-    python_type = None
+    python_type = list
 
 
 class SimpleAggregateFunction(ChSqlaType, UserDefinedType):
-    python_type = None
+    python_type = str
 
     def __init__(
         self,
@@ -532,7 +539,7 @@ class AggregateFunction(ChSqlaType, UserDefinedType):
     Note this isn't currently supported for insert/select, only table definitions
     """
 
-    python_type = None
+    python_type = str
 
     def __init__(self, *params, type_def: TypeDef = None):
         """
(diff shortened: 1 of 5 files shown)


clickhouse_connect/cc_sqlalchemy/datatypes/sqltypes.py
```diff
@@ -425,7 +429,10 @@ def __init__(self, element: ChSqlaType | type[ChSqlaType] = None, type_def: Type
             if callable(element):
                 element = element()
             type_def = TypeDef(values=(element.name,))
-        super().__init__(type_def)
+        ChSqlaType.__init__(self, type_def)
+        # Set item_type directly; calling ARRAY.__init__ would reject nested Array(Array(T)),
+        # which CH supports natively (CH expresses dimensions via nesting, not a dim count).
+        self.item_type = sqla_type_from_name(type_def.values[0])
```
joe-clickhouse, line 436: This comment is correct, however I did some testing and without the `__init__` calling something like `session.execute(select(Item.tags)).unique()` fails with `'Array' object has no attribute 'as_tuple'`. I traced it through and the short story is that I think we need to set `self.as_tuple = False` here as well.

clickhouse_connect/cc_sqlalchemy/inspector.py
```diff
@@ -135,12 +176,15 @@ def reflect_table(
         else:
             reflected_columns = self.get_columns(table.name, schema)
 
+        pk_columns = set(get_pk_constraint(self.bind, table.name, schema)["constrained_columns"])
```
joe-clickhouse: Remove

clickhouse_connect/cc_sqlalchemy/inspector.py
```diff
@@ -135,12 +176,15 @@ def reflect_table(
...
         for col in reflected_columns:
             name = col.pop("name")
             if (include_columns and name not in include_columns) or (exclude_columns and name in exclude_columns):
                 continue
             col_type = col.pop("type")
             col_args = {key: value for key, value in col.items() if value is not None}
+            if name in pk_columns:
+                col_args["primary_key"] = True
```
joe-clickhouse: Remove

clickhouse_connect/cc_sqlalchemy/inspector.py
```diff
@@ -119,6 +119,47 @@ def get_dictionary_metadata(connection, table_name: str, schema: str | None = No
...
 
+def get_pk_constraint(connection, table_name: str, schema: str | None = None) -> dict[str, Any]:
+    database = _database_name(connection, schema)
+    result_set = connection.execute(
+        text(
+            "SELECT name FROM system.columns WHERE database = :database AND table = :table_name AND is_in_primary_key = 1 ORDER BY position"
+        ),
+        {"database": database, "table_name": table_name},
+    )
+    return {"constrained_columns": [row.name for row in result_set], "name": None}
```
joe-clickhouse: Remove

clickhouse_connect/cc_sqlalchemy/dialect.py
```diff
@@ -89,11 +89,14 @@ def get_table_names(self, connection, schema=None, **kw):
             cmd += " FROM " + quote_identifier(schema)
         return [row.name for row in connection.execute(text(cmd))]
 
+    def get_columns(self, connection, table_name, schema=None, **kw):
+        return get_columns(connection, table_name, schema)
+
     def get_primary_keys(self, connection, table_name, schema=None, **kw):
-        return []
+        return get_pk_constraint(connection, table_name, schema)["constrained_columns"]
```
joe-clickhouse: Remove

clickhouse_connect/cc_sqlalchemy/dialect.py
```diff
@@ -89,11 +89,14 @@ def get_table_names(self, connection, schema=None, **kw):
...
 
     def get_pk_constraint(self, connection, table_name, schema=None, **kw):
-        return {"constrained_columns": [], "name": None}
+        return get_pk_constraint(connection, table_name, schema)
```
joe-clickhouse: Revert back to `{"constrained_columns": [], "name": None}`

joe-clickhouse (commented):
Hi @coby-astsec thanks for this! Appreciate the work. A few things:

1. Dialect-level `get_columns` and the shared module-level extraction are a good fix for `MetaData.reflect()` and `get_multi_columns` which would otherwise fail.
2. I like the concrete `python_type` compatibility fix as well.
3. Rebasing `Array` on `sqlalchemy.types.ARRAY` is useful too, but I think we need to set `self.as_tuple = False` in the definition. I put a comment in the code.

Then as far as the primary key stuff goes, I'm going to request that we cut it and we revert back to

```python
    def get_primary_keys(self, connection, table_name, schema=None, **kw):
        return []

    def get_pk_constraint(self, connection, table_name, schema=None, **kw):
        return {"constrained_columns": [], "name": None}
```

The reason is because `PRIMARY KEY` in ClickHouse doesn't guarantee uniqueness even if specified. And if it's not specified it defaults from `ORDER BY`. The point is I think the identity assertion should come from application code, not from default dialect reflection. 

So `MetaData.reflect()` & SQLAlchemy core will work fine without a PK once `get_columns()` is implemented. However, SQLAlchemy ORM does need a logical identity key and we're in luck because SQLAlchemy explicitly lets users provide one even if the database does not declare or enforce it. E.g.

```python
events = Table(
    "events",
    metadata,
    Column("tenant_id", UInt64, primary_key=True),
    Column("event_id", UInt64, primary_key=True),
    autoload_with=engine,
)
```

or with ORM mapping:

```python
class Event(Base):
    __table__ = events
    __mapper_args__ = {
        "primary_key": [events.c.tenant_id, events.c.event_id]
    }
```

Again, the point is I don't think the dialect should allow defining a sparse primary index as a safe ORM identity because downstream consumers are going to assume it is. But the app developer can explicitly say that it is if that's the case.

For the record, a lot of the other clients for db's without a PK concept follow this pattern as well, e.g. `pydruid`, `pinotdb`.

Happy to discuss further or help out if needed!

clickhouse_connect/cc_sqlalchemy/datatypes/sqltypes.py
coby-astsec: Agree that we need to set `as_tuple`.

Note this does not make `.unique()` succeed on array columns.
Setting `as_tuple=False` causes CH arrays to come back as Python lists, which are not hashable, so dedup is not possible either way.
The fix just makes it fail later on: with `as_tuple = False`, `.unique()` raises SQLAlchemy's `InvalidRequestError` instead.

coby-astsec:
Thanks for the review,
Regarding the Primary Keys, I agree and have removed those changes, and concerning `as_tuple`, I implemented what you asked, but also what you tried to do wouldn't work either way - I added more detail about this as a reply to your comment. 

Let me know if there's anything else!

joe-clickhouse:
Thanks @coby-astsec! Looking good. Only things left I'd request:
1. The PR summary is now stale so let's please get that updated to reflect the actual changes
2. Optional, but I think it's worth adding a test to make sure a user-defined primary key survives reflection:

```python
def test_user_declared_primary_key(test_engine: Engine, test_db: str):
    """A user-declared primary key on a pre-declared column survives reflection."""
    common.set_setting("invalid_setting_action", "drop")
    with test_engine.begin() as conn:
        conn.execute(text(f"DROP TABLE IF EXISTS {test_db}.reflect_pk_test"))
        conn.execute(
            text(
                f"CREATE TABLE {test_db}.reflect_pk_test (org_id UInt32, id UInt64, payload String) "
                "ENGINE MergeTree ORDER BY (org_id, id)"
            )
        )

    table = db.Table(
        "reflect_pk_test",
        db.MetaData(schema=test_db),
        db.Column("org_id", UInt32, primary_key=True),
        db.Column("id", db.BigInteger, primary_key=True),
        autoload_with=test_engine,
    )
    assert [c.name for c in table.primary_key.columns] == ["org_id", "id"]
    assert {c.name for c in table.columns} == {"org_id", "id", "payload"}
```

3. Rebase

Thanks!

coby-astsec:
@joe-clickhouse all done, sorry for the delay 😄

joe-clickhouse:
@copilot resolve the merge conflicts in this pull request

joe-clickhouse (approved):
Looks good! Thanks for the contribution @coby-astsec

PR #771, opened 2026-05-30
docs: document concurrent client usage

vividbaek:
## Summary

Document how clickhouse-connect clients should be used in concurrent applications. Sharing a session-bound client across multiple threads or concurrent request handlers can raise concurrent query or session locking errors, so this PR recommends per-thread or request-local client ownership and points asyncio users to the async client.

## What changed

- Added a README section explaining why a session-bound client should not be shared concurrently.
- Added a runnable example at `examples/concurrency_examples.py`.
- Included three patterns:
  - thread-local client ownership
  - request-local client ownership
  - asyncio usage with `get_async_client()`
- Recommended closing clients explicitly when they are no longer needed.

## Why

Sharing session state such as temporary tables, session settings, or a session_id across concurrent queries can cause conflicts and confusing production failures. The new docs call out safe ownership patterns more clearly.

## Verification

- Ran the example successfully in a virtual environment.
- Confirmed the script works against a local ClickHouse server.

## Notes

This is a docs-only change. No runtime behavior was modified.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,3 +91,14 @@ Then create a client with `clickhouse_connect.get_async_client()`. See the
 
 The documentation for ClickHouse Connect has moved to
 [ClickHouse Docs](https://clickhouse.com/docs/integrations/python)
+
+### Concurrency and Session Ownership
+
+Do not share a session-bound client across concurrent threads or request handlers. Each client should be owned by one thread, one request, or one coroutine at a time, especially when using temporary tables, session settings, or an explicit session_id.
+
+Recommended patterns:
+- Threaded apps: create one client per thread.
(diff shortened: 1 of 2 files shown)


joe-clickhouse:
Hi @vividbaek! Thanks for this. A few things here, though. As it happens, we're in the process as we speak of moving the docs locally from the general docs repo. You'll see in #768 we just did this. So moving forward conceptual advice like this better fits there. In this case, it'd specifically go in docs/advanced-usage.mdx after this branch is rebased. That said, once you take a look at those docs, you'll see we do have some advice already covering this. Please take a look and if anything you think needs more thorough coverage, please adjust accordingly. Let me know if you have any questions.

PR #775, opened 2026-06-03
Add support for refreshable JWT access tokens

zvonand:
Now the Client object only accepts a static string containing a token. However, sometimes (e.g. for longer-running programs) this may not be enough. Now, you have to externally check whether an auth failure happened (i.e. when a token expired) and feed a new token to the client. After this change, the client will try to use the `token_provider` callable if authentication failed. This happens at most once: if the new token also fails -- the error will be surfaced.

## Summary
Support a callable token provider with auto-refresh on auth failure

## Checklist

### DIFF
--- a/clickhouse_connect/driver/asyncclient.py
+++ b/clickhouse_connect/driver/asyncclient.py
@@ -2035,6 +2035,8 @@ class AsyncClient(Client):
                         response.close()
                         continue
                 if self._token_provider and not auth_retried and response.headers.get(ex_header) == auth_failed_ex_code:
+                    if retry_body is None and not (data is None or isinstance(data, (bytes, bytearray, str, dict))):
+                        await self._error_handler(response)  # non-replayable body, surface the auth error instead of retrying
                     auth_retried = True
                     self.set_access_token(await self._resolve_token())
                     req_headers["Authorization"] = self.headers["Authorization"]
--- a/clickhouse_connect/driver/asyncclient.py
+++ b/clickhouse_connect/driver/asyncclient.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import gzip
+import inspect
 import io
 import json
 import logging
@@ -60,6 +61,7 @@ from clickhouse_connect.driver.transform import NativeTransform
 logger = logging.getLogger(__name__)
 columns_only_re = re.compile(r"LIMIT 0\s*$", re.IGNORECASE)
 ex_header = "X-ClickHouse-Exception-Code"
+auth_failed_ex_code = "516"  # ClickHouse AUTHENTICATION_FAILED
 ex_tag_header = "X-ClickHouse-Exception-Tag"
 
 if "br" in available_compression:
@@ -197,6 +199,7 @@ class AsyncClient(Client):
         password: str | None = None,
         database: str | None = None,
         access_token: str | None = None,
+        token_provider: Callable[[], str | Awaitable[str]] | None = None,
         compress: bool | str = True,
         connect_timeout: int = 10,
         send_receive_timeout: int = 300,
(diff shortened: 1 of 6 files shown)


clickhouse_connect/driver/httpclient.py
```diff
@@ -606,6 +612,14 @@ def _raw_request(
...
                     self._error_handler(response, True)
                 logger.debug("Retrying requests with status code %d", response.status)
+            elif self._token_provider and not auth_retried and response.headers.get(ex_header) == auth_failed_ex_code:
+                auth_retried = True
+                self.set_access_token(self._token_provider())
+                headers["Authorization"] = self.headers["Authorization"]
+                if retry_body is not None:
+                    kwargs["body"] = retry_body()
+                response.close()
+                logger.debug("Refreshing access token after authentication failure")
```
joe-clickhouse, line 625: This will retry after refreshing the token but it never checks if the body can actually be resent again. On inserts, an `insert_block` can be a generator or a file-like object and if the body was already consumed before the 516 comes back, the retry will send and empty or partial body which the server will accept just fine, erroneously leaving the user thinking the data was correctly inserted. We should follow the pattern just above here in the connection-error path where it retries only when `retry_body` is set or is replayable otherwise it raises the error. Something like:

```suggestion
            elif self._token_provider and not auth_retried and response.headers.get(ex_header) == auth_failed_ex_code:
                body = kwargs.get("body")
                if retry_body is None and not (body is None or isinstance(body, (bytes, bytearray, str))):
                    self._error_handler(response)  # non-replayable body, surface the auth error instead of retrying
                auth_retried = True
                self.set_access_token(self._token_provider())
                headers["Authorization"] = self.headers["Authorization"]
                if retry_body is not None:
                    kwargs["body"] = retry_body()
                response.close()
                logger.debug("Refreshing access token after authentication failure")
```

clickhouse_connect/driver/asyncclient.py
```diff
@@ -2018,6 +2034,15 @@ async def _raw_request(
...
                         continue
+                if self._token_provider and not auth_retried and response.headers.get(ex_header) == auth_failed_ex_code:
+                    auth_retried = True
+                    self.set_access_token(await self._resolve_token())
+                    req_headers["Authorization"] = self.headers["Authorization"]
+                    if retry_body is not None:
+                        data = await retry_body()
+                    logger.debug("Refreshing access token after authentication failure")
+                    response.close()
+                    continue
```
joe-clickhouse, line 2049: Same body-replay problem exists here.

```suggestion
                if self._token_provider and not auth_retried and response.headers.get(ex_header) == auth_failed_ex_code:
                    if retry_body is None and not (data is None or isinstance(data, (bytes, bytearray, str, dict))):
                        await self._error_handler(response)  # non-replayable body, surface the auth error instead of retrying
                    auth_retried = True
                    self.set_access_token(await self._resolve_token())
                    req_headers["Authorization"] = self.headers["Authorization"]
                    if retry_body is not None:
                        data = await retry_body()
                    logger.debug("Refreshing access token after authentication failure")
                    response.close()
                    continue
```

joe-clickhouse (commented):
Hi @zvonand. Looks good overall. I left a couple of inline comments around the insert retry path resending a body that may already be consumed.

One other thing worth noting in this implementation is that on the async client the token provider can be hit concurrently. E.g. if several requests are in flight when the token expires they each get a 516 and each calls the provider on its own since there's no lock around the refresh. For a cheap or local provider that's fine but for one that hits an IdP over the network, it's a burst of duplicate refreshes, and if that endpoint rate limits it could translate into failed requests. The sync client is one request per instance so it doesn't have this issue.

I think this is pretty narrow so I don't think we need to change anything here in this PR but a single `asyncio.Lock` around the refresh would coalesce it to one call if we want it later. You can do that here if you want or I think it's at least worth a line in the docstring saying the provider may be invoked concurrently and is assumed to be safe to get called that way.

zvonand:
> You can do that here if you want or I think it's at least worth a line in the docstring saying the provider may be invoked concurrently and is assumed to be safe to get called that way.

I added a note, thanks for the review!
I think the current behavior is correct: let the the user worry about his callable being parallel-safe, not us.

Also, adding a simple lock in our code can become a bottleneck in some cases (maybe user wants to handle this differently in his code?)

joe-clickhouse (approved):
@zvonand looks good. I added some unit tests and fixed an issue with the dbapi path. Thanks!

PR #776, opened 2026-06-04
Add client-level HTTP headers

river-walras:
## Summary
Add a connection-level `headers` keyword to both sync and async client creation so callers can attach HTTP headers to every request made by clickhouse-connect, including the initialization queries sent during client construction.

## Use Case
This supports ClickHouse deployments exposed through HTTP gateways that require request headers before traffic reaches ClickHouse. A concrete example is Cloudflare Tunnel with Cloudflare Access / Zero Trust service tokens:

```python
    host="clickhouse.example.com",
    secure=True,
    username="default",
    password="...",
    headers={
        "CF-Access-Client-Id": "...",
        "CF-Access-Client-Secret": "...",
    },
)
```

Per-request `transport_settings` can already send HTTP headers for normal queries, but it cannot help with client initialization because `_init_common_settings()` runs before the caller can pass per-request options. Connection-level headers close that gap.

## Parameter Design
I used an explicit `headers` parameter instead of promoting `transport_settings` to the constructor because the new values are plain HTTP headers, and `transport_settings` already has overloaded request-level meaning in the driver. Keeping the names separate makes the public API easier to discover and avoids spreading that ambiguity to client creation.

Header merge order is intentionally:

1. Driver defaults, including auth and `User-Agent`
2. Client-level `headers`
3. Per-request `transport_settings`
4. Forced `Host` from `server_host_name`

This lets users intentionally override driver-generated headers when they need a gateway-specific escape hatch, while preserving per-request overrides for one-off calls.

## Tests
- Added unit coverage for sync client headers being present during initialization.
- Added sync and async unit coverage for request-level override behavior.
- Added shared client factory coverage with Cloudflare-style headers.

Local validation run:

- `uv run ruff check clickhouse_connect/driver/__init__.py clickhouse_connect/driver/httpclient.py clickhouse_connect/driver/asyncclient.py tests/integration_tests/test_client.py tests/unit_tests/test_driver/test_httpclient.py`
- `uv run pytest tests/unit_tests/test_driver/test_httpclient.py::TestHttpClientHeaders tests/unit_tests/test_driver/test_httpclient.py::TestAsyncClientHeaders -q`

The focused integration test could not be completed locally because no ClickHouse server was listening on `localhost:8123`.

### DIFF
--- a/clickhouse_connect/driver/__init__.py
+++ b/clickhouse_connect/driver/__init__.py
@@ -132,7 +132,7 @@ def create_client(
     :param settings: ClickHouse server settings to be used with the session/every request
     :param headers: Additional HTTP headers to send with every request. This can be used for proxy or gateway
       authentication, such as Cloudflare Access service token headers. These headers are applied after driver defaults,
-      so they can intentionally override headers such as Authorization, User-Agent, or Accept-Encoding.
+      so they can intentionally override headers such as Authorization or User-Agent.
     :param generic_args: Used internally to parse DBAPI connection strings into keyword arguments and ClickHouse settings.
       It is not recommended to use this parameter externally.
 
@@ -258,7 +258,7 @@ async def create_async_client(
     :param settings: ClickHouse server settings to be used with the session/every request
     :param headers: Additional HTTP headers to send with every request. This can be used for proxy or gateway
       authentication, such as Cloudflare Access service token headers. These headers are applied after driver defaults,
-      so they can intentionally override headers such as Authorization, User-Agent, or Accept-Encoding.
+      so they can intentionally override headers such as Authorization or User-Agent.
     :param generic_args: Used internally to parse DBAPI connection strings into keyword arguments and ClickHouse settings.
       It is not recommended to use this parameter externally
     :param connector_limit: Maximum number of allowable connections to the server
--- a/clickhouse_connect/driver/__init__.py
+++ b/clickhouse_connect/driver/__init__.py
@@ -93,6 +93,7 @@ def create_client(
     secure: bool | str = False,
     dsn: str | None = None,
     settings: dict[str, Any] | None = None,
+    headers: dict[str, str] | None = None,
     generic_args: dict[str, Any] | None = None,
     **kwargs,
 ) -> Client:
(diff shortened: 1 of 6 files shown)


clickhouse_connect/driver/__init__.py
```diff
@@ -221,6 +256,9 @@ async def create_async_client(
     :param dsn: A string in standard DSN (Data Source Name) format. Other connection values (such as host or user)
       will be extracted from this string if not set otherwise.
     :param settings: ClickHouse server settings to be used with the session/every request
+    :param headers: Additional HTTP headers to send with every request. This can be used for proxy or gateway
+      authentication, such as Cloudflare Access service token headers. These headers are applied after driver defaults,
+      so they can intentionally override headers such as Authorization, User-Agent, or Accept-Encoding.
```
joe-clickhouse: The `Accept-Encoding` part here is not quite accurate. Client-level headers are applied to `self.headers`, but the normal query path later sets `Accept-Encoding` from `self.compression`, then per-request `transport_settings` can still override it which I think is the right call. Request-specific headers should win over connection defaults. Please just drop `Accept-Encoding` from the list of headers that client-level headers can override.

```suggestion
      so they can intentionally override headers such as Authorization or User-Agent.
```

clickhouse_connect/driver/__init__.py
```diff
@@ -115,6 +130,9 @@ def create_client(
...
```
joe-clickhouse: See the other comment related to this.

```suggestion
      so they can intentionally override headers such as Authorization or User-Agent.
```

joe-clickhouse (commented):
Hi @river-walras thanks for the contribution and thanks for addressing the comments so far. I have two things on my end:

1. Sync `ping()` currently bypasses the header merge in `_raw_request()` because it calls `self.http.request(...)` directly in `httpclient.py`. Please build the ping headers from the client headers, e.g. `headers=dict_copy(self.headers)` so that ping has them as well. Note the async `ping()` already sends them through the session defaults so we're good there. This will restore parity. A unit test for sync ping with client headers would be good to add.
2. I left a note in the code about Accept-Encoding in the docstrings.

Otherwise we're about there!

joe-clickhouse (approved):
Looks good, thanks @river-walras!

PR #778, opened 2026-06-04
Add Mintlify docs verification workflow

Blargian:
## Summary

Verify this repo's `docs/` as a slice of the aggregated ClickHouse docs site over on ClickHouse/ClickHouse. The job runs inside the clickhouse/docs-builder image, where git, python3 and mint (mintlify CLI) are all present, so nothing is installed and this repo vendors no scripts.

It fetches the check driver from ClickHouse/ClickHouse at runtime; the driver shallow/sparse-clones the aggregator docs and any scripts needed to run the docs check, replaces the language client docs copy in the aggregator repo with this repo's `docs/` (from the branch the docs check runs on) and runs the same checks that ClickHouse/ClickHouse docs check runs from the docs root. A non-zero exit fails the job.

Requires:
1. https://github.com/ClickHouse/ClickHouse/pull/106258 (refactors the Mintlify docs check as part of praktika to be easily consumable from other repos without needing praktika as a dependency)
2. PR to add the Mintlify site in https://github.com/ClickHouse/mintlify-docs-dev to the server repo.
## Checklist


.github/workflows/docs_verify.yml
```diff
@@ -0,0 +1,46 @@
...
+  verify:
+    runs-on: ubuntu-latest
+    name: Mintlify checks
+    # Run inside the docs image so every dependency (git, python3, mint, node)
+    # is present -- nothing is installed, and this repo vendors no scripts.
+    container:
+      image: clickhouse/docs-builder
+    steps:
+      - name: Checkout clickhouse-connect
+        uses: actions/checkout@v4
```
joe-clickhouse: We'll need to update this to v5

```suggestion
        uses: actions/checkout@v5
```

joe-clickhouse:
According to Claude:

Status update after https://github.com/ClickHouse/ClickHouse/pull/106258 merged today. I re-ran the failed Mintlify check and the workflow itself now works end to end. The driver downloads, splices this repo's docs into the aggregator, and runs both checks.

The remaining failures are not problems with this workflow.

1. The aggregator docs on ClickHouse master are still a skeleton. The missing snippet and most of the broken links are pages that exist in mintlify-docs-dev but have not been imported to the server repo yet. The check cannot pass until that lands.
2. Our docs still used the old /core link prefix, which does not exist in the new site layout. Fix is up in #794. Once that merges, a re-run here will pick it up through the merge ref.

Blargian:
Hi @joe-clickhouse, we're now deploying the new site from `/docs` in `ClickHouse/ClickHouse`. After #794 the check here should pass.

Quite slow to run - 16 minutes (we have about 16 000 pages in the docs with translation), but I can take this to Mintlify to speed up.

joe-clickhouse (approved):
@Blargian looks like it worked! 14 min (which is what it took in this run) for the docs workflow is a bit of a deal breaker though. So the sooner we can bring that up with mintlify the better.

Blargian:
> @Blargian looks like it worked! 14 min (which is what it took in this run) for the docs workflow is a bit of a deal breaker though. So the sooner we can bring that up with mintlify the better.

Opened https://github.com/ClickHouse/ClickHouse/pull/109585 to address. Had Claude take a look at what `mint validate` does internally, and just build a bare minimum version of that which will run only for the pages in this repo, rather than on all 16 000 pages of docs (with translations). Should be good enough for our purposes - we just need to check that these pages have valid MDX and that the nav points to files which exist.  We will run the full check only when PRing the changes against ClickHouse/ClickHouse in any case.

Any chance I could ask for a review on https://github.com/ClickHouse/ClickHouse/pull/109585? 🙏

joe-clickhouse:
@Blargian amazing! Mintlify check when from 14 min to 1 min. Nice work!

PR #780, opened 2026-06-05
Add more test writing instructions to AGENTS.md

alex-clickhouse:
Ran into some issues when testing agentic bug fixes, hopefully this will lead the agent to behave a bit better :)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -113,8 +113,14 @@ When you reconcile the sub-agent's findings into your reply to the user, preserv
 - Limit parentheses.
 - Use single spaces between sentences.
 
-## Test Data
+## Writing Tests
 
+- A bug fix needs a regression test. Write tests that FAIL on the unpatched code and PASS with your fix, and confirm both directions before claiming the fix works.
+- If existing tests break due to your changes, do not edit the tests to make them pass, unless the tests were asserting fundamentally incorrect behavior. If that was the case, call it out explicitly. If you cannot justify the new expected value from the issue or from `Server Behavior Is Authoritative`, your code is wrong, not the test.
+- Derive expected values from the spec and from empirically testing against the server.
+- Inspect existing tests covering similar areas to make sure you are writing tests in the right style and with the right coverage.
+- Make sure to test both the happy path as well as sad paths. Make sure to cover all relevant edge cases.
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -113,8 +113,15 @@ When you reconcile the sub-agent's findings into your reply to the user, preserv
...
+- If existing tests break due to your changes, do not edit the tests to make them pass, unless the tests were asserting fundamentally incorrect behavior. If that was the case, call it out explicitly. If you cannot justify the new expected value from the issue or from `Server Behavior Is Authoritative`, your code is wrong, not the test.
+- Derive expected values from the spec and from empirically testing against the server.
+- Inspect existing tests covering similar areas to make sure you are writing tests in the right style and with the right coverage.
+- Make sure to test both the happy path as well as sad paths. Make sure to cover all relevant edge cases.
+- When testing types, cover the full type matrix: ClickHouse type hints compose and formatting is recursive, so a change to how a value is formatted must be tested across all the shapes it can take: scalar, Array(T), Tuple(...), Array(Tuple(...)),  Nullable(T), and spelling and case variants of the type name.
+- Aim to be complete, but also terse. Don't use two tests to cover what could be done in one.
```
mshustov, line 124: Do we have any noteworthy patterns for parametrized tests that we should point out?

joe-clickhouse: Yes, I'll leave in a separate comment.

joe-clickhouse (approved):
Looks good! Regarding the test matrix stuff, I'd add one calling out how to paramaterize a test.

- Express a matrix of cases with `@pytest.mark.parametrize` rather than near-duplicate test bodies. This satisfies the terseness goal below and is the right place for the type-shape and case-variant coverage above.

And then another specifically bringing up the fact that sync and async client can use a paramaterized fixture to test both one a single test.

- Client behavior must hold for both the sync and async clients. For integration tests use the `param_client` and `call` fixtures from `tests/integration_tests/conftest.py`, which run one test body against both transports via the `client_mode` parameter. See `tests/integration_tests/test_temporal.py` for the pattern. Do not write a sync-only test for a change that touches shared client code.

alex-clickhouse:
Don't merge this yet pls, I'm running some experiments and they'd be contaminated if we changed AGENTS.md mid-way through.

joe-clickhouse:
@alex-clickhouse just let me know when you're ready to merge.

PR #792, opened 2026-06-10
fix uuid and ip address bind parameters rendering unquoted inside containers

### LINKED ISSUE #791: `format_bind_value` doesn't quote UUID/IPv4Address/IPv6Address inside Array
### Describe the bug
`format_query_value` (client-side substitution) handles `uuid.UUID`, `IPv4Address`, and `IPv6Address` by quoting them. `format_bind_value` (server-side params) has no branch for these types, so they fall through to return `str(value)`.

At top level this is harmless (a bare string is correct for a `{name:String}` param). But nested inside an Array/Tuple/Map, the element is emitted unquoted, producing an invalid literal e.g. an IN-list of UUIDs bound to `{name:Array(String)}` becomes `[019e1780-…, …]` and the server rejects it with `Code: 26 ... Cannot parse quoted string ... cannot be parsed as Array(String)`.

This is wrong when using SQLAlchemy `server_side_params`, where `Column.in_([...])` of UUID/IP objects compiles to an `Array(String)` bind.

### Steps to reproduce

```python
import uuid
from clickhouse_connect.driver.binding import format_bind_value, format_query_value

u = uuid.UUID("019e1780-3b41-7673-a645-17f9b60fe8ec")

format_bind_value([u], top_level=False)   # [019e1780-3b41-7673-a645-17f9b60fe8ec]   <-- unquoted, invalid
format_query_value([u])                    # ['019e1780-3b41-7673-a645-17f9b60fe8ec'] <-- correct
```

### Expected behaviour

`format_bind_value` should quote UUID/IPv4Address/IPv6Address at nested level, matching `format_query_value`, so container literals are valid.

#### Environment
* clickhouse-connect version: `1.2.0`

Giuzzilla:
## Summary
`format_bind_value` now quotes `uuid.UUID`, `IPv4Address`, and `IPv6Address` at nested level, matching `format_query_value`, so Array/Tuple/Map literals bound to server-side params (e.g. SQLAlchemy Column.in_ of UUIDs) are valid.

Closes #791 

## Checklist


joe-clickhouse (approved):
Looks good. thanks!

PR #799, opened 2026-06-12
add mypy to CI with a per-module baseline for issue #692

joe-clickhouse:
## Summary
Sets up the type checking foundation for #692 so annotation work can proceed in small PRs.

- Adds mypy 2.1.0, pinned in tests/test_requirements.txt like ruff
- Adds a Mypy job to CI that runs alongside the Ruff job
- Adds `[tool.mypy]` config in `pyproject.toml` targeting Python 3.10
- Ignores missing imports for optional dependencies without stubs and for the compiled `driverc` modules
- Baselines the modules that do not yet pass with `ignore_errors` so CI is green from the start
- Contributors annotate a module, remove it from the baseline list, and CI enforces it stays clean
- `py.typed` gets restored in a final PR once the baseline list is empty

### DIFF
--- a/.github/workflows/on_push.yml
+++ b/.github/workflows/on_push.yml
@@ -44,6 +44,22 @@ jobs:
       - name: Ruff lint check
         run: ruff check clickhouse_connect tests examples setup.py
 
+  typecheck:
+    runs-on: ubuntu-latest
+    name: Mypy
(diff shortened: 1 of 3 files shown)


peter-leonov-ch (approved):
Types ❤️

Issue #802, opened 2026-06-16
IncompleteRead exception causes successful but incomplete read.


Viktor-Nordell:
### Describe the bug

The `ResponseSource` class contains the function `buffered` which may silently swallow an exception resulting in an incomplete read. Very bad as the client thinks that all data was read, even if it actual failed in the middle of the read.

### Steps to reproduce
The following scenario results in the issue in `buffered()`, in `httputil.py`:
1. One or more chucks are read successfully.
2. Read error during `next(read_gen, None)`.
3. `read_error` is set.
4. `if not chunk` is true, `done = True`.
5. The successfully read chucks are drained.
6. `if read_error and not data_received` is `false`, since a previous chuck was read successfully.
7. `read_error` is never raised. 

### Expected behaviour

The code should raise an exception if a read error occurs. 

### clickhouse-connect and/or ClickHouse server logs

```
unexpected failure to read next chunk
Traceback (most recent call last):
  File "/usr/local/lib/python3.13/site-packages/urllib3/response.py", line 905, in _error_catcher
    yield
  File "/usr/local/lib/python3.13/site-packages/urllib3/response.py", line 1434, in read_chunked
    chunk = self._handle_chunk(amt)
  File "/usr/local/lib/python3.13/site-packages/urllib3/response.py", line 1374, in _handle_chunk
    self._fp._safe_read(2)  # type: ignore[union-attr] # Toss the CRLF at the end of the chunk.
    ~~~~~~~~~~~~~~~~~~~^^^
  File "/usr/local/lib/python3.13/http/client.py", line 652, in _safe_read
    raise IncompleteRead(data, amt - len(data))
http.client.IncompleteRead: IncompleteRead(0 bytes read, 2 more expected)

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/lib/python3.13/site-packages/clickhouse_connect/driver/httputil.py", line 247, in buffered
    chunk = next(read_gen, None)  # Always try to read at least one chunk if there are any left
  File "/usr/local/lib/python3.13/site-packages/urllib3/response.py", line 1260, in stream
    yield from self.read_chunked(amt, decode_content=decode_content)
  File "/usr/local/lib/python3.13/site-packages/urllib3/response.py", line 1407, in read_chunked
    with self._error_catcher():
         ~~~~~~~~~~~~~~~~~~~^^
  File "/usr/local/lib/python3.13/contextlib.py", line 162, in __exit__
[...]

PR #803, opened 2026-06-16
Always raise exception on read error

Viktor-Nordell:
## Summary
This will make sure that any read error exception are always raised, even if there is an error after the first successful read.
Fix for https://github.com/ClickHouse/clickhouse-connect/issues/802

## Checklist


joe-clickhouse:
Hi @Viktor-Nordell, thanks! The fix itself seems ok but there are a few things that need addressing first:

1. `test_network_error_after_data_received_does_not_raise` (in `tests/unit_tests/test_driver/test_httputil.py`) asserts the old swallow behavior. Flip it to assert that partial data is yielded and then `OperationalError` is raised with the `ConnectionError` chained as the cause. In the current Pr, this test fails.
2. Please update the CHANGELOG
3. There is now a stale comment at `httputil.py:249` still says "if no data was received," which no longer applies. Please trim it.

I'm out this week so likely won't be able to merge until next week but if you're able to work on this in the meantime, that'd be appreciated. Thanks again!

Viktor-Nordell:
Thanks for the quick response @joe-clickhouse! 
Please take a look again, all three issues should be fixed.

joe-clickhouse (approved):
@Viktor-Nordell thanks! Look good. Pushed a few minor changes. Thanks for the contribution!

joe-clickhouse (approved):
Looks good. Thanks for the contribution @Viktor-Nordell!

PR #804, opened 2026-06-17
refactor(datatypes): add type annotations to pass mypy (part of #692)

rikettsie:
## Summary

This second PR continues the effort of annotating (issue #692), ensuring mypy check is clean on `datatypes` module:

- Remove all clickhouse_connect.datatypes.* entries from the mypy ignore_errors baseline;
- Fix LSP violations in `_data_size`: `Collection` over `Sequence`;
- Narrow `_array_type` in `IntBase`: replace `cast(str, ...)` calls with asserts;
- Replace `cast(Any, index).dtype` with a `# type: ignore[attr-defined]`;
- Document `# type: ignore` on `IPv6Address` private attribute access;
- Narrow the key type of `ch_read_formats` and `ch_write_formats` from `Any` to `type`;
- Fix `_data_size` to use integer division (`//`) to match the `int` return type;


joe-clickhouse (approved):
LGTM

PR #805, opened 2026-06-17
refactor(driver-utils): add type annotations to pass mypy (part of #692)

rikettsie:
## Summary

Third PR on annotations (part of issue #692) , targeting driver utilities.

All driver utility modules are annotated, to pass mypy cleanly.
All corresponding entries from the `ignore_errors` baseline are removed.

**Note:** three `driver/` modules remain to tackle: `client.py`, `httpclient.py`, `asyncclient.py` (50, 47, and 24 errors respectively) and this will be done in the next PR, dedicated to driver clients.

### Details

**Trivial widening / None defaults**
- `types.py`: `last_message: bytes | None`;
- `exceptions.py`: `# type: ignore[misc]` on cyclic `Warning` base;
- `query.py`, `npquery.py`, `common.py`, `transform.py`: optional ctor params and `None` defaults made explicit;

**`tzinfo | None` threading (`binding.py`)**
- `format_query_value`, `str_query_value`, `format_bind_value`, `DT64Param.format` all widen tz param to `tzinfo | None`;
- `bind_query` return type corrected to `tuple[str | bytes, dict[str, str]]` — the binary-bind path always returned `bytes`, mypy just didn't know;

**Private attribute access**
- `dataconv.py`: marked `IPv4Address._ip` with `# type: ignore[attr-defined]` with comment explaining the CPython performance trick;
- `httputil.py`: on `http.client._MAXHEADERS` the same _ignore_ marker;

**`compression.py`**
- Removed erroneous `@abstractmethod` from `Compressor.compress_block`: `Compressor` does not extend `ABC`, the method has a real body used by `null_compressor`, and `Compressor` is instantiated directly (a true abstract class would have raised `TypeError` when doing so);
- `comp_map` annotated, `get_compressor` accepts `str | None`, `isinstance` check replaces `try/except TypeError`;

**`asyncqueue.py`**
- `EOF_SENTINEL: Any` to stand in for a queued `T` value on EOF;
- `# type: ignore[union-attr]` with comments on `_loop` calls where the loop is guaranteed bound by `_bind_loop()`;

**`ddl.py`**
- Introduced `_NamedType` Protocol to avoid a circular import with `datatypes.base`; both `ClickHouseType` and `_DDLType` satisfy it structurally;

**`__init__.py`**
- eliminated the  `int | None` propagation in `port or parsed.port or 0`;
- added `username or ""` guards against `None` at the `HttpClient` boundary;
- `token_provider` widened to `Callable[[], str | Awaitable[str]] | None`.

:point_right:  **Also:** I widened `QueryContext.parameters` to `Sequence | dict[str, Any] | None` because the public `Client` methods already accept both forms and `finalize_query` (which is the consumer) already handles both. The previous `dict | None` annotation was too narrow and drove me to use too many casts in the callers (I discovered this friction when doing the PR #806).


joe-clickhouse (approved):
LG

PR #806, opened 2026-06-18
refactor(driver-clients): add annotations to client, httpclient, asyncclient (part of #692)

rikettsie:
## Summary

This is the fourth PR on annotations, part of the ongoing mypy effort tracked in issue #692:

- Annotates `client.py`, `httpclient.py`, and `asyncclient.py`;
- Removes those three modules from the `ignore_errors` baseline in `pyproject.toml`.

### Notable decisions

- `insert()` now raises `ProgrammingError("No table specified for insert") `explicitly when `table=None` and no context is provided; previously mypy was silenced with `cast(str, table)` which gave no runtime protection;
- `column_names` in `insert()` is narrowed, from `Iterable[str]` to `Sequence[str]`: `create_insert_context` uses `len()` and iterates multiple times, so `Iterable` was always the wrong type here; this is a minor breaking change for callers passing a generator, but such usage **was** already broken at runtime;
- `raw_insert()`: `BinaryIO` removed from the isinstance check: `typing.BinaryIO` is not valid for `isinstance` at runtime, so the original check was dead code; file-like objects fall through to the else branch, where the query goes as a URL param and the body is streamed as-is;
- `set_response_tz` is now guarded: the previous unconditional call with a `None` argument was a latent bug since the method signature only accepts `tzinfo`; that's covered by three new unit tests in `TestResponseTimezone` (see below notice);
- `dict_copy` source param is corrected from `dict` to `dict | None`.

### A new test added

The `set_response_tz` method previously received `None` on every response without a timezone header; the guard added in **driver/httpclient.py (lines 352-353)** made it conditional and these three tests pin that behaviour.

## Checklist


joe-clickhouse (approved):
LG

PR #807, opened 2026-06-18
refactor(cc_sqlalchemy): add type annotations to pass mypy (part of #692)

rikettsie:
## Summary

This is the final layer of the mypy annotation work tracked in issue #692.

It annotates all 9 `cc_sqlalchemy` modules so they pass mypy cleanly, removes them from the `ignore_errors` baseline in `pyproject.toml`, and restores `clickhouse_connect/py.typed` (the PEP 561 marker telling all type checkers this package **is** typed). Restoring that file is what formally closes #692.

:point_right: This PR should be merged as the last one of the series, and in particular, after #805 to "use" its changes (the `QueryContext` widening), so be rebased from `main` including it.

### Notable changes per file

- `dialect.py`, `sql/preparer.py`: one `# type: ignore[assignment]` each for a class-level attribute that intentionally shadows a SQLAlchemy base type;
- `ddl/custom.py`, `ddl/tableengine.py`: implicit `Optional` fixed (`str | None`) on all `__init__` params that default to `None`; base class `arg_names`/`quoted_args`/`optional_args`/`eng_params` given explicit types; `tuple_expr` got `expr_name: str` and `-> str`;
- `alembic/utils.py`: added `None` guards for config and script.upgrade_ops before attribute access;
- `alembic/impl.py`: `# type: ignore[union-attr]` on the driver connection chain (`connection.connection.driver_connection.client`), which is opaque to alembic stubs;
- `ddl/tableengine.py`: `TableEngine` now inherits from `SchemaItem` directly (which is `SchemaEventTarget` + `Visitable`); this removes the `# type: ignore[arg-type]` suppression in `alembic/impl.py` where engine instances are passed to `Table()`;
- `datatypes/base.py: annotated `_ch_type_cls`, `_instance_cache`, `schema_types`; removed dead `_instance` attribute; `# type: ignore[union-attr]` on `build()` and `__init__` because `_ch_type_cls` / `_instance_cache` are `None` on the base class but always set to real values by `__init_subclass__` before any instance is created (mypy cannot see that invariant);
- `sql/__init__.py`: `# type: ignore[attr-defined]` throughout for monkey-patching `Select (FINAL, SAMPLE, PREWHERE, LIMIT BY, ARRAY JOIN);` unavoidable without stub infrastructure; added `format_table -> str` and full type signatures on `prewhere`, `limit_by`, `_apply_array_join`;
- `datatypes/sqltypes.py `:
  - `# type: ignore[misc]` on all classes that inherit from both `ChSqlaType` and a SQLAlchemy type;
  - the MRO conflict is inherent to `ChSqlaType` not inheriting from `TypeEngine` directly; resolving it cleanly would require making `ChSqlaType` extend `TypeEngine`( :point_right: which is an API change left for a separate decision);
  - `TypeDef | None` widened on all `__init__` params;
  - added explicit `None` guards in `Enum`, `Array`, `Map`, `SimpleAggregateFunction` constructors (these also improve the error messages when called incorrectly);
  - `Nullable.__new__` and `LowCardinality.__new__` got return type `ChSqlaType` with `# type: ignore[misc]` because `__new__` must return a subtype of `cls` per Python's type system.


joe-clickhouse (approved):
Looks good. Huge thanks @rikettsie for the work on this! Really appreciate the time and effort.

rikettsie:
Thanks!

PR #814, opened 2026-06-23
chore: add local mypy pre-commit hook mirroring CI

rikettsie:
## Summary

Adds a local pre-commit hook that runs mypy (the same bare invocation as the CI "Mypy check" step in `on_push.yml`).
Also adds a one-line mention in `CONTRIBUTING.md`'s **Style Guide** section, mirroring the existing ruff instructions.


joe-clickhouse (approved):
Looks good. Pushed a small change to not run mypy unless .py files are changed.

PR #826, opened 2026-06-29
Return None from first_item/first_row on empty result

### LINKED ISSUE #824: first_item and first_row crash on empty result instead of returning None
These methods should safely return `None` when result set is empty. Currently they just crash with `IndexError`, making them useless — you still have to manually check `row_count` before using them.

### Expected behaviour

Both `first_item` and `first_row` should return `None` when query returns no rows, following the same pattern as `dict.get()` and ORM `.first()` methods in other libraries (SQLAlchemy, Django, etc.).

### Code example

```python

client = get_client(host='clickhouse')
client.command("CREATE TABLE IF NOT EXISTS test (id Int32)")
result = client.query("SELECT * FROM test WHERE id = 999")  # empty result

item = result.first_item  # should return None
row = result.first_row    # should return None
```

### Current behaviour

```bash
Traceback (most recent call last):
  File "<python-input-4>", line 6, in <module>
    item = result.first_item
           ^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.14/site-packages/clickhouse_connect/driver/query.py", line 389, in first_item
    return dict(zip(self.column_names, self.result_set[0]))
                                       ~~~~~~~~~~~~~~~^^^
IndexError: list index out of range
```

### Version

clickhouse-connect==1.3.0

arpitjain099:
first_item and first_row throw an IndexError on an empty result set instead of returning None, which is what the docstrings (and #824) say should happen. Both properties were just indexing element [0] without checking whether any rows came back, so any query that returns nothing crashes the caller. I added a row_count == 0 guard to each so they short-circuit and return None in both the row-oriented and column-oriented cases. There's a new unit test in test_native_read.py covering the empty result (None) and the non-empty control, plus a CHANGELOG note. Closes #824.


joe-clickhouse (approved):
Looks good. Thanks @arpitjain099!

PR #854, opened 2026-07-07
SQLAlchemy: use SHOW DATABASES in has_database() to include DataLakeCatalog databases (closes #849)

### LINKED ISSUE #849: SQLAlchemy dialect: has_database() uses system.databases and misses DataLakeCatalog databases
## Description

The SQLAlchemy dialect's `has_database()` classmethod queries `system.databases` to determine whether a database exists:

https://github.com/ClickHouse/clickhouse-connect/blob/main/clickhouse_connect/cc_sqlalchemy/dialect.py#L106-L108

```python
@staticmethod
def has_database(connection, db_name):
    return (connection.execute(text(f"SELECT name FROM system.databases WHERE name = {format_str(db_name)}"))).rowcount > 0
```

Databases created with the `DataLakeCatalog` engine (Iceberg / Unity / Glue catalogs) are **not** listed in `system.databases` unless the session/profile setting `show_data_lake_catalogs_in_system_tables=1` is enabled. As a result, `has_database()` returns `False` for an existing DataLakeCatalog database, causing SQLAlchemy schema-existence checks to be incorrect.

`SHOW DATABASES` returns these databases regardless of that setting, so it is the more reliable source. Note that the sibling method `get_schema_names()` (the direct analog of JDBC `getSchemas()`) already uses `SHOW DATABASES` and is therefore correct — only `has_database()` is affected.

This is the clickhouse-connect analog of the clickhouse-java issue where JDBC `DatabaseMetaData.getSchemas()` used `system.databases` and omitted DataLakeCatalog databases.

## ClickHouse server version

26.6.1.1193 (code analysis only; the behavioral difference cannot be reproduced here without an external data-lake catalog configured, but it follows directly from `system.databases` not exposing `DataLakeCatalog` databases by default).

## Reproduction

```python
from sqlalchemy import create_engine, text
from clickhouse_connect.cc_sqlalchemy.dialect import ClickHouseDialect

engine = create_engine("clickhousedb://localhost:8123")

# Assume a DataLakeCatalog database "my_lake" exists on the server and
# show_data_lake_catalogs_in_system_tables is NOT set in the session/profile.

with engine.connect() as conn:
    # get_schema_names uses SHOW DATABASES -> correctly includes "my_lake"
    names = ClickHouseDialect().get_schema_names(conn)
    assert "my_lake" in names            # passes

    # has_database uses system.databases -> misses "my_lake"
    exists = ClickHouseDialect.has_database(conn, "my_lake")
    # Expected: True
    # Actual:   False   <-- bug
    assert exists
```

Expected: `has_database()` reports `True` for an existing DataLakeCatalog database.
Actual: `has_database()` reports `False` because `system.databases` omits it.

## Suggested fix

In `clickhouse_connect/cc_sqlalchemy/dialect.py`, `has_database()` should not rely on `system.databases`. Options:
- Check membership against `SHOW DATABASES` (consistent with `get_schema_names()`), or
- Use `EXISTS DATABASE <name>`, or
- Add `show_data_lake_catalogs_in_system_tables=1` to the existence query.

## Link

Relayed from ClickHouse/clickhouse-java#2907

steps-re:
Closes #849.

`ClickHouseDialect.has_database()` determined database existence via `system.databases`:

```python
connection.execute(text(f"SELECT name FROM system.databases WHERE name = {format_str(db_name)}")).rowcount > 0
```

Databases created with the `DataLakeCatalog` engine (Iceberg / Unity / Glue) are **not** listed in `system.databases` unless `show_data_lake_catalogs_in_system_tables=1` is set, so `has_database()` returned `False` for databases that actually exist, breaking SQLAlchemy schema-existence checks.

`SHOW DATABASES` returns those databases regardless of that setting, and the sibling `get_schema_names()` already uses it — so this just makes `has_database()` consistent with it:

```python
db_name in [row.name for row in connection.execute(text("SHOW DATABASES"))]
```

`format_str` was only used here, so I dropped it from the import. Added a mocked-connection unit test (`test_has_database.py`) that asserts membership behaviour and that `SHOW DATABASES` is used rather than `system.databases` (no server needed). Local run: the new test plus the rest of `tests/unit_tests/test_sqlalchemy/` (232 tests) pass; `ruff` clean.


joe-clickhouse (approved):
@steps-re Thanks for this. In general, it looks good. I confirmed the bug against the server source. Turns out `system.databases` hid `DataLakeCatalog` and other remote databases by default on servers 25.10 through 26.4 and it was fixed server side in 26.5, so this did genuinely break `has_database()` on those versions.

To you PR, I made one change. I switched the implementation from `SHOW DATABASES` to `EXISTS DATABASE`, which in this case I think is more idiomatic, matches the sibling `has_table()`, keeps the check server side, and is correct on every version instead of depending on the full database listing. I also added the changelog entry you were missing.

Issue #856, opened 2026-07-08
SQLAlchemy DDL Alias and Materialized columns can't have comments


ElPincheTopo:
### Describe the bug
When using SQLAlchemy to create a table, materialized and alias columns silently drop the comment option.

### Steps to reproduce
1. Define a SQLAlchemy table with a materialized or alias column.
2. Use sqlalchemy to create the table.
3. Look at the table definition in clickhouse cli. The column comment will be missing.

### Expected behaviour
The column comment should be included.

### Code example
```python
Column(
    'my_col',
    DateTime64(3),
    clickhouse_materialized=text('now64(3)'),
    comment='This is a column comment'
)
```

### clickhouse-connect and/or ClickHouse server logs
clickhouse-connect==1.4.2
sqlalchemy==2.0.51
#### Environment
* clickhouse-connect version:
* Python version: 3.12.12
* Operating system: Ubuntu 22.04

#### ClickHouse server
* ClickHouse Server version: 25.8.28

joe-clickhouse:
Closed by #857

joe-clickhouse:
Hi @ElPincheTopo, the fix to this issue was included in release [v1.5.0](https://github.com/ClickHouse/clickhouse-connect/releases/tag/v1.5.0). Thanks!

PR #857, opened 2026-07-08
Refactor SQLalchemy DDL compiler to process defaults after materialized and alias

ElPincheTopo:
Removed early return statements for materialized and alias options to allow further processing. These were preventing materialized and alias columns from having comments, ttls and codecs.

https://github.com/ClickHouse/clickhouse-connect/issues/856

## Checklist


joe-clickhouse (approved):
Hey @ElPincheTopo thanks. Looks good. I pushed a small follow up. The column clauses now emit in the order ClickHouse requires, `COMMENT` then `CODEC` then `TTL`, since a column with both a codec and a comment was otherwise rejected by the server. I also kept `DEFAULT`, `MATERIALIZED`, and `ALIAS` mutually exclusive and added a changelog entry.

PR #859, opened 2026-07-08
Add docs sync workflow to ClickHouse/ClickHouse

Blargian:
## What

Adds `.github/workflows/docs_sync.yml`, a self-contained, parameterized workflow that opens (or refreshes) a PR on `ClickHouse/ClickHouse` replacing `docs/integrations/language-clients/python` with this repo's `docs/` folder — in the spirit of ClickHouse/ClickHouse#109746. This repo is the source of truth; the sync is one-way.

The sync PR is labeled `pr-autogenerated-docs`, which the aggregator's docs check uses to allow the sync PR to make updates (TO DO). Ordinarily, changes made to `docs/integrations/language-clients/python` in ClickHouse/ClickHouse will have a failing docs check which points the contributor to this repo for any changes they want to make.

## Triggers

- **Merged PR with the `sync-docs` label** → syncs immediately, without waiting for a release (the expedited path).
- **Published release** → syncs, gated by `RELEASE_SCOPE` (default `major` = X.0.0 only; configurable to `minor`/`all`; prereleases skipped).
- **Manual dispatch** → forces a sync.

## How

- A stable branch (`robot/docs-sync-clickhouse-connect`) is force-pushed each run, so there is always a single open PR that is one commit off the target branch.
- The target docs dir is wiped before copying, so deletions and renames propagate (same semantics as `docs_verify.yml`).
- Cross-repo auth reuses the existing GitHub App pattern from `cross-repo-bug-relay.yml` (`actions/create-github-app-token@v3` + `WORKFLOW_AUTH_PUBLIC_*` secrets).
- All configuration lives in the top-level `env:` block for easy reuse in another client repo.

## Prerequisites before this is functional

- [x] Create the `sync-docs` label on this repo.
- [ ] Confirm the GitHub App behind `WORKFLOW_AUTH_PUBLIC_APP_ID` is installed on `ClickHouse/ClickHouse` with `contents: write` + `pull-requests: write`.


Blargian:
Hi @joe-clickhouse, I'll need your help adding the `sync-docs` label. There are a few small pieces on ClickHouse/ClickHouse for the `pr-autogenerated-docs` label (introduced in https://github.com/ClickHouse/ClickHouse/pull/109441) and the logic around disallowing edits to the copies [check_readonly_copies.py)](https://github.com/ClickHouse/ClickHouse/blob/master/ci/jobs/scripts/docs/check_readonly_copies.py) and [readonly_copies.json)](https://github.com/ClickHouse/ClickHouse/blob/master/ci/jobs/scripts/docs/readonly_copies.json) (which I see now has a duplicate in it), but we can start testing the sync flow from this PR already - no need to wait for that.

joe-clickhouse:
@Blargian sounds good! I just created the `sync-docs` label.

.github/workflows/docs_sync.yml
```diff
@@ -0,0 +1,199 @@
...
+  # This repo's docs folder (the source of truth).
+  SOURCE_DOCS_PATH: docs
+  # Stable branch on TARGET_REPO, force-pushed each run.
+  SYNC_BRANCH: robot/docs-sync-clickhouse-connect
+  # Label the sync PR gets on TARGET_REPO.
+  PR_LABEL: pr-autogenerated-docs
+  # Label on a merged PR in THIS repo that triggers the expedited path.
+  SYNC_LABEL: sync-docs
+  # Which releases trigger a sync: major (X.0.0), minor (X.Y.0), or all.
+  RELEASE_SCOPE: major
```
joe-clickhouse: I think this should be "all". I'd expect a docs sync under any release, not just major versions bumps as those are very rare.

Blargian: Makes sense, updated. I was thinking more ahead to when we introduce the versioning. The aggregator would place the docs into the appropriate folder depending on the release version:

```
/integrations/language-clients/python
├── 1_4/
│   ├── doc1.mdx
│   └── doc2.mdx
├── 1_3/
│   ├── doc1.mdx
│   └── doc2.mdx
├── 1_2/
│   └── ...
└── 1_1/
    └── ...
```

Creating a full copy of the docs for each minor version could explode page count quickly if we're doing this for all client repos - which has a performance cost for the site. I'll give some thought to how to handle this.

joe-clickhouse: Ah I see...yeah that'd probably be too much. Is it possible to just create new folders per major release, but still have the docs update each merge then within that major? So V1 is the version you choose on the docs page, but it's always kept up to date?

Blargian: Yep, that would work nicely! Or we could have multiple sets for each minor within the most recent major, and replace them with a single set of docs once we bump to the next major.

joe-clickhouse: Ah ok yeah, that makes sense too. I know there won't be many docs changes between patch releases so that does seem too frequent. Minors usually will have a few updates though. Do we need to revert this now? I don't want to create a bunch of unnecessary document bloat prematurely. Or is it all ok for now and we can revisit when we start implementing the versioning?

Blargian: Good for now, we can update this again when Mintlify is ready with the sidebar versioning functionality for us.

joe-clickhouse (approved):

PR #872, opened 2026-07-15
Refactor: implement pluggable backend architecture

joe-clickhouse:
## Summary

The sync and async clients each carried their own copy of the request building, the transport handling, and the multi-step init handshake, written directly against `urllib3` and `aiohttp`. Supporting any non-HTTP execution engine meant either duplicating the client surface per backend or reimplementing internals that were never designed as an interface. Issue discussions around chDB support such as #753 ran into exactly this as will a future TCP transport implementation.

This PR restructures the driver internals around a small backend layer behind the existing clients, then validates the design with an experimental in-process chDB backend. The public API is unchanged.

**Backend layer**
- `driver/backend/contracts.py` defines the `SyncBackend` and `AsyncBackend` protocols. There is a `Capabilities` declaration plus the typed execute methods:
  - `execute_query`
  - `execute_command`
  - `execute_data_insert`
  - `execute_raw_insert`
  - `execute_raw_query`
  - `execute_raw_stream`
- and the typed lifecycle methods:
  - `ping`
  - `close`
  - `close_connections`
- Request building is now shared. Pure planner functions in `driver/backend/httpcommon.py` build the params, headers, and payload for every request kind, used identically by both HTTP transports and pinned by unit tests.
- `HttpSyncBackend` and `HttpAsyncBackend` own pooling, session lifecycle, the retry and auth refresh loop, ping, and close.
- The init handshake and insert-context discovery are written once as generators in `driver/backend/orchestration.py` and driven by both clients, replacing line-for-line sync/async duplicates.
- `SyncBackendClient` in `driver/backendclient.py` implements the semantic client methods once against the backend contract. `HttpClient` extends it and keeps construction, HTTP compat properties, and settings storage.

**chDB backend**
- `get_client(interface='chdb')` or a `chdb://` DSN returns a standard client backed by an embedded chDB engine, covering the full query, insert, streaming, and Arrow surface through the shared Native codec. Installable with `pip install clickhouse-connect[chdb]`, marked experimental.
- chDB allows one engine per process, so concurrent clients share a refcounted session with locking rules that prevent database and settings cross-talk. No async client and no external data support, both raise.

**Compatibility**

- The refactor is held to zero observable behavior change. Requests are byte identical except one normalization in that async requests now emit URL query parameters in the same order as sync.
- Any deliberate behavior changes are listed in the CHANGELOG.
- Undocumented re-exports were removed from `driver.httpclient`, `driver.client`, and `driver.asyncclient`. These names were never canonical there and should be imported from their home modules.
- The published `chdb` 4.2.1 `cc_backend`, which subclasses `Client` directly, runs unmodified against this branch.

## Checklist


clickhouse_connect/driver/_backend/chdb_backend.py
```diff
@@ -0,0 +1,793 @@
...
+
+    def __init__(self, conn: Any):
+        self.conn = conn
+        self.lock = threading.Lock()
+        self.refcount = 0
+        self.active_database: str | None = None
+        self.stream_owner: int | None = None
+
+
+# chdb permits one active engine per process: a second chdb.connect() with a
```
ShawnChen-Sirius: The 4.0.0 EmbeddedServer refactor changed the model, the latest implementation is:
 - multiple connects to the same path coexist; close() only releases its own handle
 - concurrent queries are safe, on one handle or across handles
 - USE and SET are per-handle
 - stream on handle A + query on handle B works; two concurrent streams work
 - the one hazard left: a query on the same handle while its stream will silently returns an empty result

So I recommand give each ChdbClient its own chdb.connect() handle instead of sharing one. You keep a per-client lock for the stream hazard, but the global lock, the per-call USE rebind, and the SET snapshot/restore dance all go away — along with the "settings visible across all clients" limitation in the doc, and the current behavior where one client's open stream blocks every other client until it's drained.

clickhouse_connect/driver/_backend/chdb_backend.py
```diff
@@ -0,0 +1,793 @@
...
+# different path raises (or deadlocks on some versions), and a second connect
+# on the same path deadlocks inside the C extension. Backends therefore share
+# one refcounted _EngineSession; the connection really closes when the last
+# user releases it.
+_SESSIONS: dict[str, _EngineSession] = {}
+_SESSIONS_LOCK = threading.Lock()
+
+
+def _acquire_session(conn_str: str) -> _EngineSession:
+    with _SESSIONS_LOCK:
```
ShawnChen-Sirius: Suggest normalizing the path before comparing — the current raw-string key can reject clients that point at the same directory:
· get_client(interface='chdb', path='/data/db')
· get_client(interface='chdb', path='/data/db/')   # ProgrammingError, but it's the same engine
With a normalized key the second call would just reuse the existing session.

clickhouse_connect/driver/_backend/chdb_backend.py
```diff
@@ -0,0 +1,793 @@
...
+        for other_str, other in _SESSIONS.items():
+            if other_str != conn_str and other.refcount > 0:
+                raise ProgrammingError(
+                    f"chdb allows one engine per process: close the clients using {other_str!r} before connecting to {conn_str!r}"
+                )
+        session = _SESSIONS.get(conn_str)
+        if session is None:
+            session = _EngineSession(chdb.connect(conn_str))
```
ShawnChen-Sirius: If some other code in the process already holds a chdb connection on a different path, this chdb.connect raises a raw RuntimeError through get_client. So worth catching and re-raising the potential error:
  ▎ try:
  ▎     session = _EngineSession(chdb.connect(conn_str))
  ▎ except RuntimeError as ex:
  ▎     raise ProgrammingError(f"Unable to open chdb engine: {ex}") from ex
That keeps get_client's error contract and matches the ProgrammingError your registry guard above raises for the same situation.

joe-clickhouse:
@ShawnChen-Sirius great review and suggestions! Thanks for taking the time. All of your points were valid and correct. I've implemented fixes for them. Please take another look and see what you think.

ShawnChen-Sirius:
LGTM

ShawnChen-Sirius (approved):

PR #874, opened 2026-07-16
fix: escape dict-valued settings as ClickHouse map literals (#501)

### LINKED ISSUE #501: Failure passing string filters in additional_table_filters settings param when calling .query
### Describe the bug
Passing the additional_table_filters with a string filter, fails to handle the escaping of single quotes correctly, and crashes instead, making it impossible to pass query filters that contain strings. 

### Steps to reproduce
1. Setup a Clickhouse machine 
2. Add a table with a string field
3. Perform a query passing additional_table_filters for the string with a filter

### Expected behaviour
The query will work with the filter applied 
### Code example
```python
ch = get_client(...)
ch.query("CREATE OR REPLACE TABLE my_table (
      a String
    )")
ch.query("INSERT INTO my_table (a) VALUES ('some_value');")

ch.query(
    "SELECT a from my_table", settings={"additional_table_filters": {"a":"'some_value'"} 
)
```

self = <clickhouse_connect.driver.httpclient.HttpClient object at 0x1220c6180>, response = <urllib3.response.HTTPResponse object at 0x12395f970>, retried = False

    def _error_handler(self, response: HTTPResponse, retried: bool = False) -> None:
        if self.show_clickhouse_errors:
            try:
                err_content = get_response_data(response)
            except Exception:  # pylint: disable=broad-except
                err_content = None
            finally:
                response.close()
    
            err_str = f'HTTPDriver for {self.url} returned response code {response.status}'
            err_code = response.headers.get(ex_header)
            if err_code:
                err_str = f'HTTPDriver for {self.url} received ClickHouse error code {err_code}'
            if err_content:
                err_msg = common.format_error(err_content.decode(errors='backslashreplace'))
                if err_msg.startswith('Code'):
                    err_str = f'{err_str}\n {err_msg}'
        else:
            err_str = 'The ClickHouse server returned an error.'
    
>       raise OperationalError(err_str) if retried else DatabaseError(err_str) from None
E       clickhouse_connect.driver.exceptions.DatabaseError: HTTPDriver for http://localhost:8123 received ClickHouse error code 26
E        Code: 26. DB::Exception: Cannot parse quoted string: expected opening quote ''', got '"'. (CANNOT_PARSE_QUOTED_STRING) (version 25.4.1.2934 (official build))

../.venv/lib/python3.12/site-packages/clickhouse_connect/driver/httpclient.py:395: DatabaseError

#### Environment
* clickhouse-connect version: 0.8.15
* Python version: 3.12.11
* Operating system: MacOS Sonoma 14.5

#### ClickHouse server
* ClickHouse Server version: 25.4.1
Any table with string field. 
CREATE OR REPLACE TABLE my_table (
      a String
    )
INSERT INTO my_table (a) VALUES ('some_value');

zerafachris:
Fixes #501

Root cause: `additional_table_filters` (and any Map-typed setting) passed via
`query(settings={...})` crashed with `DB::Exception: Cannot parse quoted string:
expected opening quote ''', got '"'`. `_validate_settings()` stored the raw Python
dict instead of a validated string, so `urlencode()` fell back to Python's own
`str()`/`repr()` of the dict — which mixes single and double quotes — producing
invalid ClickHouse map-literal syntax.

What changed: `_validate_setting()` now renders dict values as a proper
single-quoted, escaped ClickHouse map literal (via the existing
`binding.str_query_value` helper), independent of the unrelated
`dict_parameter_format` setting which only governs bound query parameters.
`_validate_settings()` now forwards that rendered string for dict/list/tuple
values instead of the raw object, so it survives `urlencode()` unchanged.
Scalar setting values (int/bool/str) are unaffected — same behavior as before.

Test evidence:
- Added `TestQuery.test_raw_query_with_map_setting` (tests/unit_tests/test_driver/test_httpclient.py).
  Before fix: `AssertionError: assert False == isinstance({'a': "'some_value'"}, str)`.
  After fix: passes, asserting `additional_table_filters` renders as
  `{'a': '\'some_value\''}` with no double quotes.
- Verified end-to-end against a live ClickHouse 25.6 server: before the fix, the
  exact repro from the issue raises `Code: 26 ... Cannot parse quoted string`;
  after the fix, `SELECT ... SETTINGS additional_table_filters=...` succeeds.
- Full suite: `pytest tests/unit_tests` (922 passed, 1 skipped),
  `pytest tests/integration_tests/test_client.py` (84 passed),
  `ruff format --check`, `ruff check`, and `mypy` all clean.

CHANGELOG.md updated under UNRELEASED per repo convention (AGENTS.md).

Prepared with AI assistance (Claude, agent: claude-sonnet-5), reviewed for correctness before submission.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## UNRELEASED
 
+### Bug Fixes
+- Dict-valued settings such as `additional_table_filters` no longer crash with `DB::Exception: Cannot parse quoted string` when passed through `query()`'s `settings` parameter. The value was rendered with Python's own `str()`/`repr()` of the dict, which mixes single and double quotes and is not valid ClickHouse map-literal syntax; it is now rendered as a properly single-quoted, escaped ClickHouse map literal. Closes [#501](https://github.com/ClickHouse/clickhouse-connect/issues/501).
+
 ## 1.5.0, 2026-07-15
 
 ### Bug Fixes
--- a/clickhouse_connect/driver/client.py
+++ b/clickhouse_connect/driver/client.py
@@ -19,7 +19,7 @@ from clickhouse_connect.datatypes import dynamic as dynamic_module
 from clickhouse_connect.datatypes.base import ClickHouseType
 from clickhouse_connect.datatypes.registry import get_from_name
 from clickhouse_connect.driver import options, tzutil
-from clickhouse_connect.driver.binding import quote_identifier
+from clickhouse_connect.driver.binding import quote_identifier, str_query_value
 from clickhouse_connect.driver.common import (
     StreamContext,
     coerce_bool,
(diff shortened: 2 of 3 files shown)


joe-clickhouse (approved):
Hi @zerafachris thanks for the fix! Looks good to me. If you can just sign the CLA, please, I'll get this merged.

PR #875, opened 2026-07-17
Docs: move integration metadata from index.mdx comment to docs/_meta.yml

Blargian:
## Summary
Test metadata file for use with in-product documentation

Replaces the JSON metadata blob embedded as an MDX comment in `docs/index.mdx` (added in #873) with a dedicated `docs/_meta.yml` file.

- Adds `docs/_meta.yml` with the integration metadata (`schemaVersion: 1`) for website usage
- Removes the now-redundant JSON comment block from `docs/index.mdx`

Notes:
- `quickStartCode` still points at `/examples/run_async.py`, unchanged from the previous JSON comment. The path resolves in the repo.
- Volatile fields from the old JSON blob (`stars`, `version`, `packageRegistryURL`, `changelog`) are intentionally not carried over.

### DIFF
--- a/docs/_meta.yml
+++ b/docs/_meta.yml
@@ -13,7 +13,7 @@ formats: [Native, RowBinary, Arrow, Parquet, JSONEachRow, CSV, TSV]
 ecosystem: [Apache Arrow, Pandas, SQLAlchemy, Alembic, Superset]
 install:
   pip: pip install clickhouse-connect
-quickStartCode: /examples/run_async.py
+quickStartCode: /examples/quickstart.py
 desc: >-
   HTTP client with NumPy/Pandas/Arrow/Polars integration and async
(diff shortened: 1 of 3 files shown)


joe-clickhouse (approved):
@Blargian Looks good. Seeing this made me realize that we didn't actually have a super simple quick start example. The async one originally pointed to was a bit advanced for a "quick start". So I added an actually simple example and pointed to `quickStartCode` reference to it.

joe-clickhouse:
@Blargian did something else change somewhere too? The docs check is taking forever again.

Blargian:
> @Blargian did something else change somewhere too? The docs check is taking forever again.

Yep, caused the slow Mintlify check to get run. Fix here https://github.com/ClickHouse/ClickHouse/pull/110912

Issue #128, opened 2023-02-14
Kerberos Auth


tabarincev:
Hello, thank you for this great library!
Is it planned to support a Kerberos Auth with requests-kerberos?

PR #887, opened 2026-07-21
Add Kerberos Auth (#128)

pferate:
## Summary
This adds a Kerberos connection options that generates a SPNEGO ("Negotiate") token from the Kerberos credential cache for each HTTP request.

Adds use_kerberos and kerberos_hostname_override parameters to both the sync (http) and async clients, backed by a new kerberos.py module built on pyspnego/gssapi/krb5 (new kerberos extra). Because ClickHouse authenticates each HTTP request independently rather than caching auth for a session, a fresh token is generated per request rather than once at connect time.

Issue: https://github.com/ClickHouse/clickhouse-connect/issues/128

## Checklist


joe-clickhouse:
Hi @pferate thanks for this PR as well.

Before we commit to reviewing and maintaining this long term, could you share more about the environment driving it? Kerberos is [excluded](https://clickhouse.com/docs/operations/external-authenticators/kerberos) from ClickHouse Cloud, so this'll be a self-managed-only feature. Some of the other client maintainers are looking at the same question so your answers could help inform that broader decision. Specifically:

1. Is this an Active Directory realm or MIT Kerberos, and what kind of environment is it?
2. Is there an org actively running kerberized ClickHouse behind this? If so roughly at what scale?
3. Is authenticating from the process credential cache sufficient for your case?
4. Are your clients Linux only or do you also need Windows via SSPI?
5. Is there a load balancer or proxy between clients and ClickHouse in your setup?

If you're more comfortable answering these outside of a public channel, let me know.

pferate:
Hi @joe-clickhouse Personally, I don't have a dog in this race, so I can't give you any useful answers to those questions.

I was just looking through open issues and seeing where I can help contribute.  I saw that https://github.com/ClickHouse/clickhouse-connect/issues/128 was open and stale, so I picked it.

joe-clickhouse:
@pferate understood! Thanks so much for the context.

joe-clickhouse:
Hey @pferate, after working through this a bit more I've decided to close the PR without merging for now. But I really appreciate all the effort and work you put into getting it here.

The original issue is ~3.5 years old and hasn't seen any activity. The finished implementation adds >1K lines across several dozen files, along with additional client complexity, an optional dependency, and a dedicated CI job. Without any demand it's hard to justify the maintenance and complexity burden of another auth. This has nothing to do with quality, but added complexity without recent demand.

I'm also planning broader simplification of the client API in the near future. That might be a better time to revisit this kind of broader auth support and decide how it should fit into the client.

But thanks again for taking this on! The work is still available in the closed PR if/when we decide to revisit it later.

For anyone else reading this who is interested in having kerberos auth support, please upvote or add some context to #128.

PR #905, opened 2026-07-29
feat(sqlalchemy): support materialized CTEs (WITH ... AS MATERIALIZED)

### LINKED ISSUE #900: SQLAlchemy: support materialized CTEs (WITH ... AS MATERIALIZED)
### Description

We are migrating a production ClickHouse SQLAlchemy codebase to `clickhouse-connect` and hit a gap around CTE materialization.

ClickHouse 26.3+ supports materialized common table expressions — both the `enable_materialized_cte` setting and the explicit `WITH <name> AS MATERIALIZED (...)` syntax — so a CTE referenced more than once is computed **once** instead of being inlined and re-executed on every reference.

We could not find a supported way to emit `AS MATERIALIZED` for a CTE through the SQLAlchemy dialect. `Select.cte()` in SQLAlchemy 2.0 only accepts `name`, `recursive`, and `nesting`, and the `cc_sqlalchemy` dialect renders `MATERIALIZED` only for column definitions (DEFAULT/MATERIALIZED/ALIAS), not for CTEs.

### Why this matters

A common late-materialization pattern ranks rows in a narrow CTE, then references that CTE twice — once to fan out to the wide row via a JOIN, and once as an `IN (SELECT … FROM cte)` predicate that gives the planner a primary-key `KeyCondition` so the outer scan prunes instead of streaming the whole table:

```sql
WITH ranked AS (
    SELECT book_id, row_number() OVER (ORDER BY score DESC) AS result_rank
    FROM book
    WHERE genre = 'sci-fi'
    ORDER BY score DESC
    LIMIT 100
)
SELECT book.*, ranked.result_rank
FROM book ANY INNER JOIN ranked ON book.book_id = ranked.book_id
WHERE book.book_id IN (SELECT book_id FROM ranked)   -- prune hint (book_id is the ORDER BY key)
```

Because `ranked` is referenced twice and inlined by default, its (expensive) body runs twice. Marking it `AS MATERIALIZED` computes it once. Measured on a large table (~13M rows, ClickHouse 26.4.1):

| | rows read | elapsed |
| -- | --: | --: |
| default (inlined, re-executed) | 27.9M | 3.26s |
| `WITH ranked AS MATERIALIZED (…)` | **15.25M** | **2.12s** |

Note: this needs **both** the `AS MATERIALIZED` keyword **and** the `enable_materialized_cte=1` session setting — neither alone materializes the CTE in our tests (keyword-only and setting-only both leave it inlined; both together materialize it). So the per-query setting we can already pass is not enough on its own — we also need to emit the keyword, which is what this issue asks for.

### Desired API

We don't have a strong preference on the exact shape — anything that lets us emit `AS MATERIALIZED` for a CTE works. Two options that would feel natural:

```python
ranked = narrow_query.cte("ranked", materialized=True)
# or
ranked = narrow_query.cte("ranked").materialized()
```

both rendering:

```sql
WITH ranked AS MATERIALIZED ( … )
```

If you have a cleaner API in mind (e.g. a different keyword, or mirroring `AS NOT MATERIALIZED` too), that's completely fine by us — we're not attached to a specific form.

### Question

Is there already a recommended way to emit `AS MATERIALIZED` for a CTE with the current dialect? If not, would adding a materialization hint to the `cc_sqlalchemy` `Select.cte()` path be a good fit?

xoelop:
Closes #900.

ClickHouse inlines a CTE by default, so one referenced more than once has its body executed per reference. 26.3 added `WITH <name> AS MATERIALIZED (...)`, but the dialect had no way to emit it: SQLAlchemy's `Select.cte()` only takes `name`, `recursive`, and `nesting`.

```python
from clickhouse_connect.cc_sqlalchemy import cte, select

ranked = select(book.c.book_id, ...).where(...).cte("ranked", materialized=True)

# same options for a statement built with the standard sqlalchemy.select
ranked = cte(sa_select(book.c.book_id), "ranked", materialized=True)
```

The module-level form takes the statement as its first argument, matching the existing `final()` / `prewhere()` / `limit_by()` helpers, and accepts any `HasCTE` so the dialect's `Values` CTE support keeps working.

## Implementation

The keyword is attached as a CTE prefix scoped to the `clickhousedb` dialect. SQLAlchemy renders CTE prefixes between the name and the body, exactly where ClickHouse expects the keyword, so no `visit_cte` override is needed, and a statement shared with another backend compiles unchanged there.

`AS NOT MATERIALIZED` is not implemented, the server rejects it as a syntax error.

## The keyword alone is a silent no-op

The server also needs `enable_materialized_cte`, and with it off it **ignores the keyword rather than erroring**. Checked on 26.4.1: HTTP 200, no exception, nothing in `system.warnings`, and the plan simply inlines the body:

```
-- enable_materialized_cte=0
Aggregating
  Union
    Expression (...) -> ReadFromSystemNumbers
    Expression (...) -> ReadFromSystemNumbers      -- body read twice

-- enable_materialized_cte=1
MaterializingCTEs (Materialize CTEs before main query execution)
  Aggregating
    Union
      Expression (...) -> ReadFromMemoryStorage
      Expression (...) -> ReadFromMemoryStorage
  MaterializingCTE (Materializing CTE: r) -> ReadFromSystemNumbers   -- body read once
```

So forgetting the setting costs performance without failing, which the docs now call out explicitly and an integration test pins. The API deliberately does not set it for you; it goes through the normal `execution_options(settings=...)` path.

Integration tests are gated on `min_version("26.3")`.

## Tests

- `tests/unit_tests/test_sqlalchemy/test_materialized_cte.py`: rendering, default-off, the module-level form on a plain `sqlalchemy.select` and on a `Values` construct, `TypeError` on a statement without CTE support, no leakage to other dialects, the issue's join + `IN` pattern, distinct compiled-statement cache keys, and `recursive` + `materialized` together.
- `tests/integration_tests/test_sqlalchemy/test_materialized_cte.py`: runs the issue's late-materialization pattern against a real server, parametrized materialized/not, plus the silent-no-op case above.
- Extended `tests/type_check/sqlalchemy_select_smoke.py` with both entry points.

Docs and `CHANGELOG.md` updated. Ran locally: unit suite, `tests/integration_tests/test_sqlalchemy` against 26.4.5, `ruff`, `mypy`, and the consumer-install checks (both `--strict` smoke tests, plus the pyright ratchet, which reports the same count as an untouched `main`).

https://claude.ai/code/session_01VRFJAmiTn7AVS6jaeovP6u


joe-clickhouse (approved):
@xoelop thanks for the contribution! Everything pretty much looks good. I applied a few follow-ups:
 
1. I added a client-side check that rejects `recursive=True` with `materialized=True` since the server does not support recursive materialized CTEs.
2. `Values.cte()` requires SQLAlchemy 2.0.42 or later so I added a checkin the tests.
3. I added the new rendering tests to SQLAlchemy 1.4 CI.
4. I documented the experimental setting, analyzer requirement, recursive restriction, and SQLAlchemy version boundary.

Thanks again!

PR #922, opened 2026-08-01
fix(dbapi): add PEP 249 Binary; bind Time as integer seconds

### LINKED ISSUE #919: Two Core types cannot be bound through the DBAPI: LargeBinary (no PEP 249 Binary constructor) and Time
**Versions**

- clickhouse-connect 1.6.0
- SQLAlchemy 2.0.51
- Python 3.12.9, macOS
- Server: `clickhouse/clickhouse-server:25.3`

Two Core types can't be bound through the driver. Neither is a ClickHouse limitation — the database handles both fine — so I'm reporting them together, though the causes are different.

## 1. `LargeBinary` — the PEP 249 `Binary` constructor is missing

`clickhouse_connect.dbapi` doesn't define `Binary`. SQLAlchemy's bind processor calls it and gets:

```
AttributeError: module 'clickhouse_connect.dbapi' has no attribute 'Binary'
```

before any statement is sent. PEP 249 lists `Binary` among the constructors a module must supply, alongside `Date`, `Time`, `Timestamp` and the rest.

ClickHouse stores binary perfectly well — `String` holds arbitrary bytes — so this is purely the DBAPI layer being incomplete.

## 2. `Time` — the column is created and then refuses every value

`TIME` is a real ClickHouse type, stored as an integer number of seconds, and the column gets created without complaint. But a Python `time` reaches the server as the bare literal `14:30:00` where an `Int64` is expected, and the insert fails to parse.

A column that exists and rejects every value is a slightly worse outcome than one that couldn't be created, because it looks fine until you write to it. I've had to declare `Time` unstorable on this dialect for that reason.

## Suggested fixes

For the first: add `Binary` to `clickhouse_connect.dbapi` — it can be as thin as `Binary = bytes`, which is what several other drivers do.

For the second: convert a `datetime.time` to seconds since midnight in the bind path, matching the storage the server expects.

Happy to test either against my setup.

sankalpsthakur:
## Summary

Fixes #919.

Two DB-API / SQLAlchemy Core bind gaps:

1. **`Binary` missing** — SQLAlchemy `LargeBinary` calls `dialect.dbapi.Binary(value)` and raised `AttributeError` because `clickhouse_connect.dbapi` did not define it. Add `Binary = bytes` plus the other standard PEP 249 constructors (`Date`, `Time`, `Timestamp`, `*FromTicks`).

2. **`Time` binds as `HH:MM:SS`** — ClickHouse `Time`/`Time64` store integer ticks (seconds for `Time`). `format_query_value` / `format_bind_value` left `datetime.time` and `timedelta` as bare string forms that the server rejects for Int32/Int64. Convert them to integer seconds so inserts and binds succeed.

## Checklist

## Tests
- `tests/unit_tests/test_driver/test_dbapi_types.py` — `Binary` / PEP 249 constructors
- `tests/unit_tests/test_driver/test_params.py` — time/timedelta format_query_value, format_bind_value, finalize_query, server-side bind

```
pytest tests/unit_tests/test_driver/test_params.py \
       tests/unit_tests/test_driver/test_dbapi_types.py \
       tests/unit_tests/test_driver/test_temporal.py \
       tests/unit_tests/test_driver/test_binding.py -q
# 138 passed
```

## AI disclosure
This change was implemented with AI assistance (Grok). I reviewed the diff, aligned it with the issue suggestions and existing bind/temporal conventions, and ran the focused unit suite above.

### AI/LLM disclosure
- AI coding tools (including Grok and/or Codex agent-assisted editing) were used to help draft or modify code and this PR description.
- I reviewed the complete change, understand the reasoning, and ran the reported local tests before submitting.
- This submission is original work of authorship under the project CLA / contributor terms; AI output was not pasted unreviewed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Bug Fixes
 
+- DB API: add the PEP 249 `Binary` constructor (`Binary = bytes`) so SQLAlchemy `LargeBinary` bind processing no longer raises `AttributeError: module 'clickhouse_connect.dbapi' has no attribute 'Binary'`. Also expose the standard `Date`, `Time`, `Timestamp`, and `*FromTicks` constructors. Query and server-side binds of `datetime.time` and `timedelta` values now send integer seconds (ticks) instead of the bare `HH:MM:SS` string, matching ClickHouse `Time`/`Time64` storage. Closes [#919](https://github.com/ClickHouse/clickhouse-connect/issues/919).
 - Parsing a nested `Variant`, `Tuple`, `Nested`, or typed `JSON` column type whose element is an `Enum` with an escaped single quote in a value name no longer corrupts the escape sequence and fails while re-parsing the element type. Closes [#878](https://github.com/ClickHouse/clickhouse-connect/issues/878).
 - `None` nested inside an `Array` or `Tuple`, or inside a `Map` when `dict_parameter_format="map"`, now renders as the SQL `NULL` keyword instead of the `\N` sentinel used for top-level values. Top-level scalar `None` binds are unchanged. Closes [#879](https://github.com/ClickHouse/clickhouse-connect/issues/879).
 - Inserting empty bytes `b""` into a non-nullable `FixedString(N)` column now zero-pads to N bytes instead of raising `DataError`, matching the existing string and nullable-bytes write paths. Closes [#880](https://github.com/ClickHouse/clickhouse-connect/issues/880).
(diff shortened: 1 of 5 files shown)


joe-clickhouse:
Hi @sankalpsthakur, thanks for the work here but I am closing this PR because it combines two unrelated changes. The DB API constructor fix has been re-authored as a single PR with tests and a changelog entry. The `time` and `timedelta` binding changes are excluded because they corrupt `Time64` values and break working server-side binds. That work will be handled separately.

PR #923, opened 2026-08-01
fix: protect SQLAlchemy reflection from String bytes format

### LINKED ISSUE #920: Global bytes read format (set_default_formats("String", "bytes")) breaks SQLAlchemy reflection: get_columns/reflect raise TypeError, get_table_names returns bytes
## Description

`set_default_formats("String", "bytes")` (equivalently `set_read_format("String", "bytes")`) is a global read format that makes `String` columns decode to `byte[]`-style `bytes` instead of `str`. The core driver already protects itself from this: `clickhouse_connect/driver/client.py:78` defines

```python
# Orchestration queries are internal, so their decode must not be affected by
# user-configured global read formats such as set_default_formats("String", "bytes").
_INTERNAL_QUERY_FORMATS = {"String": "string"}
```

and every orchestration `QueryOp` (including the insert path's `DESCRIBE TABLE` at `driver/_backend/orchestration.py:182`) is run with that override. Inserts are unaffected — verified.

**The `cc_sqlalchemy` metadata/reflection layer has no such protection.** It issues its own `DESCRIBE TABLE` / `system.tables` / `system.columns` queries through the ordinary DBAPI cursor, so under the global bytes format every metadata cell comes back as `bytes` and the string handling downstream breaks. Two distinct failure shapes:

1. **Loud** — `Inspector.get_columns()` and any `Table(..., autoload_with=conn)` reflection raise `TypeError: a bytes-like object is required, not 'str'` at `cc_sqlalchemy/inspector.py:131`:
   ```python
   sqla_type = sqla_type_from_name(row.type.replace("\n", ""))
   ```
   `row.type` is `bytes`, and `bytes.replace` will not take `str` arguments. Note that fixing only this line moves the failure downstream — `row.name`, `row.comment`, `row.default_type`, `row.codec_expression` etc. are all `bytes` too, and `default_type == "DEFAULT"` silently evaluates `False`.

2. **Silent** — `Inspector.get_table_names()` returns `[b'my_table']` instead of `['my_table']`. No exception, and bytes table names flow onward into name comparisons and `full_table()` string building. `get_table_metadata(...).engine == "Dictionary"` in `inspector.py:125` is the same shape: `b'Dictionary' == 'Dictionary'` is `False`, so a Dictionary table would silently take the ordinary-table branch.

This is the Python analog of ClickHouse/clickhouse-cs#486 item 1 (`ReadStringsAsByteArrays` breaking `GetSchema("Columns")`): a driver-internal metadata query being decoded with a user-facing read format that was only ever meant to apply to user data. Item 2 of that issue (`GetString()` returning `"System.Byte[]"`) has no analog here — the Python DB-API has no typed string accessor — and the write-path mirror is already handled: `binding.py:248` routes `bytes`/`bytearray` through `escape_bytes` rather than `str()`.

## ClickHouse server version

26.7.1.1315 (verified at runtime, local HTTP interface).

## Reproduction

```python
import clickhouse_connect
from clickhouse_connect.datatypes.format import set_default_formats, clear_all_formats
from sqlalchemy import MetaData, Table, create_engine, inspect

client.command("DROP TABLE IF EXISTS scratch_bytes_fmt")
client.command("CREATE TABLE scratch_bytes_fmt (k UInt32, s String) ENGINE MergeTree ORDER BY k")

set_default_formats("String", "bytes")   # <- the only difference between pass and fail

conn = create_engine("clickhousedb://localhost:8123/default").connect()
insp = inspect(conn)

print("get_table_names:", insp.get_table_names()[:1])   # expect ['scratch_bytes_fmt']
print("get_columns:", insp.get_columns("scratch_bytes_fmt"))
Table("scratch_bytes_fmt", MetaData(), autoload_with=conn)

clear_all_formats()
```

Actual, with `set_default_formats("String", "bytes")` in place:

```
insert: OK
query value: [(b'hello',)]
get_columns: FAIL TypeError a bytes-like object is required, not 'str'
get_table_names: [b'scratch_bytes_fmt']
reflect: FAIL TypeError a bytes-like object is required, not 'str'
```

Control — identical script with that one line removed:

```
insert: OK
query value: [('hello',)]
get_columns: OK [('k', 'UInt32'), ('s', 'String')]
get_table_names: ['scratch_bytes_fmt']
reflect: OK ['k', 's']
```

Note `get_table_names` in the failing run: it does not raise, it just hands back `bytes`.

No existing test covers reflection under a non-default global read format — the `cc_sqlalchemy` tests all run with default formats.

## Suggested fix

Apply the same `_INTERNAL_QUERY_FORMATS` guard the core driver already uses to the metadata queries in `cc_sqlalchemy/inspector.py` (`get_columns`, `get_table_names`, `get_table_metadata`, `get_dictionary_*`), so the driver's own schema introspection decodes as text regardless of what the user set globally. This is the driver querying `system.tables` / `DESCRIBE TABLE` for its own metadata, not returning user data — that metadata is always text. Fixing at the query level rather than coercing at each use site also avoids the whack-a-mole noted above, where `row.type` is only the first of several `bytes` fields.

Reaching that from the SQLAlchemy layer requires threading a per-query format override through the dialect/DBAPI cursor (`cc_sqlalchemy/dialect.py`, `dbapi`), which today has no way to pass `query_formats`; that plumbing is the substance of the fix.

## Link

Relayed from ClickHouse/clickhouse-cs#486 — same root cause, different surface.

sankalpsthakur:
## Summary

Fixes #920.

`set_default_formats("String", "bytes")` (or `set_read_format("String", "bytes")`) is a global read format that makes `String` columns decode as `bytes`. The core driver already protects orchestration queries via `_INTERNAL_QUERY_FORMATS = {"String": "string"}`, but the SQLAlchemy reflection path (`cc_sqlalchemy` inspector + dialect metadata methods) issued ordinary DB-API queries and was unprotected.

Under the global bytes format this caused:

1. **Loud** — `Inspector.get_columns()` / `Table(..., autoload_with=...)` raised `TypeError: a bytes-like object is required, not 'str'` on `row.type.replace("\n", "")`.
2. **Silent** — `get_table_names()` returned `[b'my_table']` instead of `['my_table']`, and engine/comment comparisons like `row.engine == "Dictionary"` failed.

### Changes

- Thread per-query `query_formats` through `dbapi.Cursor.execute` / `executemany` and the SQLAlchemy dialect `do_execute*` methods, mirroring the existing `settings` / `execution_options` plumbing.
- Apply the same `_INTERNAL_QUERY_FORMATS` override to all SQLAlchemy metadata queries (`DESCRIBE TABLE`, `system.tables`, `SHOW TABLES` / `SHOW DATABASES`, dictionary create SQL, etc.) via `with_internal_query_formats(...)`.
- User SELECTs are unchanged and still honor the global bytes format.

## Checklist

## AI disclosure

This change was drafted with AI assistance and reviewed by the author before submission.

### AI/LLM disclosure
- AI coding tools (including Grok and/or Codex agent-assisted editing) were used to help draft or modify code and this PR description.
- I reviewed the complete change, understand the reasoning, and ran the reported local tests before submitting.
- This submission is original work of authorship under the project CLA / contributor terms; AI output was not pasted unreviewed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Bug Fixes
 
+- SQLAlchemy reflection (`get_columns`, `get_table_names`, `Table(..., autoload_with=...)`, and related inspector methods) no longer breaks when a global String read format such as `set_default_formats("String", "bytes")` is configured. Metadata queries now use the same internal `String` → `string` query format override as the core driver's orchestration path, so schema cells decode as `str` while user queries still honor the global format. Closes [#920](https://github.com/ClickHouse/clickhouse-connect/issues/920).
 - Parsing a nested `Variant`, `Tuple`, `Nested`, or typed `JSON` column type whose element is an `Enum` with an escaped single quote in a value name no longer corrupts the escape sequence and fails while re-parsing the element type. Closes [#878](https://github.com/ClickHouse/clickhouse-connect/issues/878).
 - `None` nested inside an `Array` or `Tuple`, or inside a `Map` when `dict_parameter_format="map"`, now renders as the SQL `NULL` keyword instead of the `\N` sentinel used for top-level values. Top-level scalar `None` binds are unchanged. Closes [#879](https://github.com/ClickHouse/clickhouse-connect/issues/879).
 - Inserting empty bytes `b""` into a non-nullable `FixedString(N)` column now zero-pads to N bytes instead of raising `DataError`, matching the existing string and nullable-bytes write paths. Closes [#880](https://github.com/ClickHouse/clickhouse-connect/issues/880).
--- a/clickhouse_connect/cc_sqlalchemy/dialect.py
+++ b/clickhouse_connect/cc_sqlalchemy/dialect.py
@@ -7,7 +7,12 @@ from sqlalchemy.exc import NoResultFound, NoSuchTableError
 
 from clickhouse_connect import dbapi
 from clickhouse_connect.cc_sqlalchemy import dialect_name, ischema_names
-from clickhouse_connect.cc_sqlalchemy.inspector import ChInspector, get_columns, get_table_metadata
+from clickhouse_connect.cc_sqlalchemy.inspector import (
+    ChInspector,
+    get_columns,
+    get_table_metadata,
+    with_internal_query_formats,
+)
 from clickhouse_connect.cc_sqlalchemy.sql import full_table
 from clickhouse_connect.cc_sqlalchemy.sql.compiler import ChStatementCompiler
 from clickhouse_connect.cc_sqlalchemy.sql.ddlcompiler import ChDDLCompiler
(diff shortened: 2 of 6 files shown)


sankalpsthakur:
Superseded by #935, same change rebased onto current `main` (resolves the `cursor.py` conflict with #927; keeps its `_leading_keyword` check and best-effort metadata probe while threading `query_formats` through). Closing this one; please review there.

PR #937, opened 2026-08-04
Add show_clickhouse_errors="scrub" to hide host and version

royxact:
## Summary
- Extends `show_clickhouse_errors` with `"scrub"`: keeps the SQL exception text and symbolic name (e.g. `UNKNOWN_TABLE`) while stripping the request URL and trailing `(version ...)` trailer from client exception messages.
- When error detail is disabled (`False`), the request URL is no longer appended to the generic error string.
- Addresses the middle ground requested in #344 (full detail vs. a generic message with no SQL error).

### Behavior

| Mode | SQL error | Symbolic name | Version | Host/URL |
|------|-----------|---------------|---------|----------|
| `True` (default) | yes | yes | yes | yes |
| `"scrub"` | yes | yes | scrubbed | hidden |
| `False` | no | no | — | hidden |

Example:

```python
client = clickhouse_connect.get_client(..., show_clickhouse_errors="scrub")
```

Verified end-to-end with Superset SQL Lab against a live ClickHouse server using `connect_args={"show_clickhouse_errors": "scrub"}`.

## Test plan
- [x] Unit tests for `scrub_error_details`, `coerce_show_clickhouse_errors`, and `build_http_error`
- [x] HttpClient/AsyncClient error-handler coverage for scrub and disabled modes
- [x] Local docker-compose Superset + ClickHouse e2e (SQL Lab `UNKNOWN_TABLE`, no URL/version leaks)
- [x] Sign CLA if prompted

Made with [Cursor](https://cursor.com)


royxact:
Retriggering CLA check after rewriting commit authors to `royxact`.

joe-clickhouse (approved):
Hey @royxact thanks for this! Looks good from my end and I think this is a great idea. The approach is I kept your core design. We pushed a follow-up commit on top of your work with the following additions:

- Typed the setting as `bool | Literal["scrub"]` and fixed the resulting mypy errors on the base class attribute and the client property overrides.
- The runtime property setters now coerce values the same way the constructor does, and unrecognized values raise `ProgrammingError` instead of disabling error detail.
- Transport failures no longer embed the `urllib3` or `aiohttp` exception text unless the mode is `True`, since those messages contain the host and URL. The full detail is logged at warning level and the original exception remains on `__cause__`.
- Mid-stream errors raised as `StreamFailureError` now honor the setting too, for both HTTP clients and chdb. Previously they carried the raw server text in every mode.
- Scrubbing is applied before `max_error_size` truncation so a partially truncated version trailer can't survive in scrub mode.
- Unified the generic disabled-mode message across the HTTP and chdb clients.
- Added integration test coverage and updated the docs and changelog.

Behavior with `True` is unchanged.

PR #948, opened 2026-08-11
Recognize $ in query parameter names

### LINKED ISSUE #936: Parameters: $ in a parameter name is not recognized by external_bind_re, so server-side binding is skipped (or type hints silently ignored)
### Describe the bug

`$` is a word character in ClickHouse's lexer: it is legal inside identifiers and inside query-parameter names. Verified against the live server used below (26.7.2.59):

```
$ curl -sG localhost:8123 --data-urlencode 'query=SELECT {id$x:Int32} AS v' --data-urlencode 'param_id$x=42'
42
$ curl -sG localhost:8123 --data-urlencode 'query=SELECT {$x:Int32} AS v' --data-urlencode 'param_$x=42'
42
$ curl -sG localhost:8123 --data-urlencode 'query=SELECT {id$:Int32} AS v' --data-urlencode 'param_id$=42'
42
```

`clickhouse_connect` disagrees, because the placeholder-detection regex at `clickhouse_connect/driver/binding.py:19` uses `\w`, which excludes `$`:

```python
external_bind_re = re.compile(r"\{(\w+):([^}]+)\}")
```

`bind_query` uses that regex for two separate decisions (`binding.py:145-176`), and `$` breaks both:

**1. Whether to bind server-side at all.** If *no* placeholder matches, `bind_query` concludes the query has no server-side placeholders and falls through to the client-side `%`-formatting path (`finalize_query`). For a query whose only placeholders contain `$`, `query % params` is a no-op (there are no `%` conversions), so the query is sent verbatim with **no `param_*` values at all** and the server rejects it:

```
Code: 456. DB::Exception: Substitution `id$x` is not set. (UNKNOWN_QUERY_PARAMETER)
```

Note this is a silent misclassification, not a name-validation error — the user gets an error that says the parameter is unset even though they passed it.

**2. Which type hint applies to each parameter.** `param_types` is keyed by the regex-captured names, so in a *mixed* query (at least one `$`-free placeholder, which makes `matches` non-empty and re-enables the server-side path) the `$`-named parameter *is* sent, but `param_types.get(k)` misses it. So `_promote_datetime64` and `_extract_tz_from_type` never run for it, and the `DateTime64` precision / timezone hints are **silently dropped**:

| Query | Bound value |
|---|---|
| `SELECT {t:DateTime64(6)} AS t` | `2024-01-02 03:04:05.123456` OK |
| `SELECT {t$x:DateTime64(6)} AS t, {a:Int32} AS a` | `2024-01-02 03:04:05` microseconds lost |

This is the same class of silent truncation that #739 / #773 fixed for the `_64` suffix, reintroduced via the name character class.

### ClickHouse server version

`26.7.2.59` (local, HTTP 8123). Both failures below are observed results, not code inspection.

### Steps to reproduce

`test_dollar_param.py` at the repo root:

```python
from datetime import datetime

import clickhouse_connect
from clickhouse_connect.driver.binding import bind_query


def test_dollar_in_param_name_binds_server_side():
    # `$` is a legal word char in ClickHouse query-parameter names.
    query, params = bind_query("SELECT {id$x:Int32} AS v", {"id$x": 42})
    assert params == {"param_id$x": "42"}  # actual: {}

    assert client.query("SELECT {id$x:Int32} AS v", parameters={"id$x": 42}).result_rows == [(42,)]


def test_dollar_in_param_name_keeps_datetime64_precision():
    dt = datetime(2024, 1, 2, 3, 4, 5, 123456)
    rows = client.query(
        "SELECT {t$x:DateTime64(6)} AS t, {a:Int32} AS a",
        parameters={"t$x": dt, "a": 1},
    ).result_rows
    assert rows == [(dt, 1)]  # actual: sub-second precision truncated
```

```
$ python -m pytest test_dollar_param.py -q
FAILED test_dollar_param.py::test_dollar_in_param_name_binds_server_side
  AssertionError: assert {} == {'param_id$x': '42'}
  Right contains 1 more item: {'param_id$x': '42'}

FAILED test_dollar_param.py::test_dollar_in_param_name_keeps_datetime64_precision
  At index 0 diff: (datetime.datetime(2024, 1, 2, 3, 4, 5), 1)
                != (datetime.datetime(2024, 1, 2, 3, 4, 5, 123456), 1)
2 failed
```

Raw `bind_query` output for the leading/trailing-`$` variants, all of which the server accepts natively:

```
bind_query('SELECT {idx:Int32} AS v',  {'idx': 42})   -> ('SELECT {idx:Int32} AS v',  {'param_idx': '42'})   # ok
bind_query('SELECT {id$x:Int32} AS v', {'id$x': 42})  -> ('SELECT {id$x:Int32} AS v', {})                    # unbound
bind_query('SELECT {$x:Int32} AS v',   {'$x': 42})    -> ('SELECT {$x:Int32} AS v',   {})                    # unbound
bind_query('SELECT {id$:Int32} AS v',  {'id$': 42})   -> ('SELECT {id$:Int32} AS v',  {})                    # unbound
```

Each of the three unbound cases then fails with `Code: 456 ... Substitution ... is not set`.

### Expected behaviour

A parameter name should be delimited the way the ClickHouse lexer delimits a word, i.e. `$` counts as a word character. `{id$x:Int32}` with `parameters={"id$x": 42}` should bind `param_id$x=42` and return `42`, matching what the server already accepts over raw HTTP — and `{t$x:DateTime64(6)}` should get the same precision promotion as `{t:DateTime64(6)}`.

### Suggested fix

One-line character-class change at `clickhouse_connect/driver/binding.py:19`, e.g.:

```python
external_bind_re = re.compile(r"\{([\w$]+):([^}]+)\}")
```

Two things worth checking alongside it:

- **Binary-bind sentinel overlap.** `bind_query` treats a key matching `k.startswith("$") and k.endswith("$") and len(k) > 1` as a binary bind (`binding.py:139`), so names like `$x$` are claimed by that convention and can't route to server-side binding. Widening the regex doesn't create the ambiguity, but it makes it reachable, so the precedence deserves an explicit decision (and probably a test).
- **Unicode axis.** Python's `\w` is Unicode-aware by default, so `{idé:Int32}` currently *matches* even though the server's word lexer is ASCII-only. That's the opposite error from `$` — too broad rather than too narrow. It's mostly unreachable in practice (`param_idé` is rejected at the URI layer), but if the regex is being touched anyway, `[A-Za-z0-9_$]` would align it with the server on both axes.

### Provenance

Found while checking whether ClickHouse/clickhouse-cs#516 (same root cause — `$` excluded from the parameter-name character class — manifesting there in the ADO.NET `@name` rewriter) applies to this client. It does, at a different layer: this client has no `@name` rewriting, but its `{name:Type}` detection regex has the same too-narrow character class.

Related: https://github.com/ClickHouse/clickhouse-cs/issues/516

Vivek1106-04:
## Summary

`$` is a word character in the ClickHouse lexer and is legal inside a query parameter name, but the placeholder regex at `clickhouse_connect/driver/binding.py` used `\w`, which excludes it.

`bind_query` uses that regex for two decisions, and both broke:

1. **Whether to bind server-side at all.** With no match, `bind_query` concluded the query had no server-side placeholders and fell through to the client-side `%` formatting path. That path is a no-op for a query with no `%` conversions, so the query went out with no `param_` values and the server rejected it with `Code: 456 ... Substitution is not set`, even though the caller passed the parameter. A silent misclassification, not a name validation error.
2. **Which type hint applies.** In a mixed query, where another placeholder kept the server-side path active, a `$` named parameter was sent but `param_types` did not contain it, so `_promote_datetime64` and `_extract_tz_from_type` never ran and `DateTime64` precision and timezone hints were silently dropped. Same class of silent truncation that the `_64` suffix handling fixed.

The fix widens the character class:

```python
external_bind_re = re.compile(r"\{([\w$]+):([^}]+)\}")
```

Closes #936

## Notes on scope

Three judgment calls worth flagging, since the issue raised two of them as open questions.

**The SQLAlchemy dialect is a second site, changed deliberately.** `cc_sqlalchemy/sql/compiler.py` carries `_bind_name_re = re.compile(r"\w+\Z")` under a comment saying it mirrors the driver's `external_bind_re`. Left alone it would reject `server_side_params` bind names that the driver now binds happily, and the comment would be false. Widened to match, with the error message updated to `[A-Za-z0-9_$]`.

**The Unicode axis is deliberately left alone.** Python's `\w` is Unicode aware, so `{idé:Int32}` matches today and binds `param_idé`, even though the server's word lexer is ASCII only. That is the opposite error, too broad rather than too narrow. Narrowing to `[A-Za-z0-9_$]` would change observable behavior for a case nobody has reported, so it is out of scope here. Happy to do it in a separate PR if you want the two axes aligned.

**`$name$` precedence is unchanged.** A key matching `k.startswith("$") and k.endswith("$")` is claimed by the binary bind convention, which substitutes into the query text. Widening the regex makes that overlap reachable, so it is pinned by two tests that pass both before and after this change: the binary bind still wins, and it still works alongside a `$` named server-side parameter in the same query.

## Verification

Both directions confirmed, per `AGENTS.md`.

- Unit tests fail on unpatched code (12 in `test_params.py`, 3 in `test_server_side_params.py`) and pass with the fix.
- Integration tests run against a live ClickHouse 25.8, the floor of the CI matrix. All 8 (3 names plus the `DateTime64` precision case, each across sync and async via `param_client` and `call`) fail unpatched and pass with the fix.
- Full unit suite: 1357 -> 1374 passed, with identical pre-existing failures and errors before and after.
- Full integration suite: 626 -> 634 passed, with the identical 23 pre-existing errors before and after.
- `ruff format --check` and `ruff check` clean. `mypy` reports "Success: no issues found in 93 source files".

The two backward-compatibility guards (binary bind sentinel, and the client-side `%` path when a query has no placeholders) pass both before and after by design.

## Checklist


joe-clickhouse (approved):
Hey @Vivek1106-04 thanks for this! The original fix looks pretty good. I did push a follow-up commit on top of yours that handles some issue that the wider name class exposed, which I found mostly by testing directly against the server lexer.

- The `[\w$]+` class accepted some names the server actually rejects. E.g. ClickHouse allows `$` inside and at the start of a bare word, but a digit-led token never absorbs one, so a name like `13a$x` lexes as multiple tokens and isn't a valid placeholder. The grammar now matches the server exactly.

- Names that start and end with `$` collided with the driver's raw binary bind convention. E.g. a parameter named `$x$` would compile through SQLAlchemy but then fail with a raw `TypeError` inside `bind_query`. Binary binds are now claimed by value instead of by name, buffer values keep the existing behavior and any other value binds server side through a normal `{name:Type}` placeholder.

All in all, looks great! Appreciate the contribution.

Vivek1106-04:
@joe-clickhouse Thanks for the review.

PR #952, opened 2026-08-11
Fix sync client sending a pathless request-target through a forwarding HTTP proxy

### LINKED ISSUE #951: Sync client omits the `/` path in the request-target when using a forwarding HTTP proxy, causing HTTP 400 (async client is unaffected)
### Describe the bug

When the **synchronous** client is used with a **forwarding HTTP proxy** (`http_proxy=...` or the `HTTP_PROXY` environment variable) and no `proxy_path` is set, the request-target sent to the proxy has **no path component**:

```
GET http://host:8123?query=SELECT+1 HTTP/1.1
```

(note the missing `/` between the authority and the `?`). Many forwarding proxies reject this as a malformed request-target and return **HTTP 400** before the request ever reaches ClickHouse.

Root cause: with no `proxy_path`, the base URL is built as `http://host:8123` (no path) in `clickhouse_connect/driver/httpclient.py`, and the sync backend then concatenates the query string directly onto it in `clickhouse_connect/driver/_backend/http_sync.py`:

```python
url = f"{self.url}?{urlencode(final_params)}"   # -> http://host:8123?query=...
```

- **Without a proxy**, urllib3 uses origin-form and normalizes the empty path to `/` (`Url.request_uri`), so `GET /?query=... HTTP/1.1` is sent and it works — which is why this is invisible unless a proxy is involved.
- **With a forwarding HTTP proxy**, urllib3 forwards the absolute URL verbatim (absolute-form request-target), so the pathless URL reaches the proxy and is rejected.

The **async** client is **not affected**: its backend always appends `/` (`clickhouse_connect/driver/_backend/http_async.py:477`, `url = f"{self.url}/"`) and passes params separately, so it sends a valid `http://host:8123/?query=...`. This sync/async inconsistency is the core of the bug.

### Steps to reproduce

1. Have a forwarding HTTP proxy in front of a plain-HTTP ClickHouse endpoint (e.g. a corporate forward proxy; verified against one that returns `Request Error (invalid_request)` for pathless targets).
2. Create a **sync** client with `http_proxy=...` and **no** `proxy_path`.
3. Run any query (e.g. `SELECT 1`). It fails with HTTP 400 from the proxy. The same call via `get_async_client(...)` succeeds.

### Expected behaviour

The sync client should send a valid request-target with a `/` path,`http://host:8123/?query=...`, matching (a) the async client and (b) urllib3's own normalization for direct connections. `SELECT 1` should return `[(1,)]`.

### Code example

```python
import clickhouse_connect

# Fails: HTTP 400 "invalid_request" from the proxy
c = clickhouse_connect.get_client(
    host="clickhouse.example.com",
    port=8123,
    username="default",
    password="<redacted>",
    http_proxy="http://proxy.example.com:80",
)
print(c.query("SELECT 1").result_rows)

# Works unpatched (async backend appends "/"):
# import asyncio
# async def main():
#     ac = await clickhouse_connect.get_async_client(
#         host="clickhouse.example.com", port=8123,
#         username="default", password="<redacted>",
#         http_proxy="http://proxy.example.com:80",
#     )
#     print((await ac.query("SELECT 1")).result_rows)  # [(1,)]
# asyncio.run(main())
```

Suggested fix in `clickhouse_connect/driver/_backend/http_sync.py` — add the `/` only when the URL has no path, so an explicit `proxy_path` is left untouched:

```python
# before
url = f"{self.url}?{urlencode(final_params)}"
# after
authority_and_path = self.url.split("://", 1)[-1]
path_sep = "" if "/" in authority_and_path else "/"
url = f"{self.url}{path_sep}?{urlencode(final_params)}"
```

(The simpler `f"{self.url}/?..."` also works and matches the async backend, but it appends a trailing slash in the `proxy_path` case; the conditional form above is a strict no-op for every existing configuration.)

Proxy response returned to the client (the request never reaches ClickHouse, so there are no server-side logs):

```
clickhouse_connect.driver.exceptions.DatabaseError: HTTP driver received HTTP status 400,
server response: <HTML><HEAD><TITLE>Request Error</TITLE></HEAD><BODY>
... <big>Request Error (invalid_request)</big> ...
Your request could not be processed. Request could not be handled ...
This could be caused by a misconfiguration, or possibly a malformed request. ...
(for url http://clickhouse.example.com:8123)
```

#### Environment
* clickhouse-connect version: 1.6.0
* Python version: 3.12.11
* Operating system: macOS (arm64)

#### ClickHouse server
* ClickHouse Server version: N/A — request is rejected by the forwarding proxy before reaching ClickHouse; reproducible independent of server version.

abhisheksurve45:
Closes #951 

## Summary

The synchronous client built its request URL as `{self.url}?...` with no path component when no `proxy_path` was set, producing `http://host:8123?query=...`.

- **Direct connections** are fine because urllib3 normalizes the empty path to `/` (`Url.request_uri`), sending `GET / query=... HTTP/1.1`. This is why the missing `/` was invisible without a proxy.
- **Through a forwarding HTTP proxy** (`http_proxy` / `HTTP_PROXY`), urllib3 forwards the absolute URL verbatim as the request-target (`GET http://host:8123?query=... HTTP/1.1`). Many proxies reject that malformed, pathless target with **HTTP 400** before the request reaches ClickHouse.

The async backend was already unaffected because it always appends `/` (`http_async.py`). This is a sync/async inconsistency.

Fix: in `http_sync.py`, add the `/` **only when the URL has no path**, so an explicit `proxy_path` is left exactly as-is. This makes the sync client consistent with the async client and with urllib3's own direct-connection normalization.

Verified live against a real forwarding proxy: sync went from HTTP 400 to `SELECT 1 -> [(1,)]`; direct connections are byte-identical on the wire before and after the change.

Compatibility:

| Config | Before -> wire | After -> wire | Change |
| --- | --- | --- | --- |
| No `proxy_path`, direct | `GET /?query=...` | `GET /?query=...` | none |
| No `proxy_path`, HTTP proxy | `GET http://host:8123?...` | `GET http://host:8123/?...` | broken -> fixed |
| `proxy_path=/ch`, direct or proxy | `.../ch?...` | `.../ch?...` | none |

## Checklist


joe-clickhouse (approved):
Thanks for the PR @abhisheksurve45! Root cuase and fix look good. I pushed a few small changes.
- I reworded the comments and changelog because the old request target is technically valid HTTP, just non normalized, and some proxies reject it or drop the query string.
- I moved the normalization into `__init__` so it's computed once instead of per request. - I also reused the existing test helpers and noted the async trailing slash divergence, which is now tracked in #963.

PR #959, opened 2026-08-12
fix: escape backslashes in SQLAlchemy DDL comment rendering

bunlongheng:
### Summary

`ClickHouseDDLHelper.render_comment` escaped a comment string by doubling only the
single quote:

```python
escaped = comment.replace("'", "''")
return f"'{escaped}'"
```

The inherited SQLAlchemy literal renderer used the same quote-only strategy for generic
strings compiled with `literal_binds=True` and for string column clauses.

ClickHouse treats the backslash as an escape character inside string literals, so quote
doubling alone is not safe. A backslash before a quote consumes the first quote of the
doubled pair. The second quote then closes the literal, and the remaining text is parsed
as SQL.

### Impact

User-controlled strings can reach SQL literals generated by the SQLAlchemy dialect and
Alembic. A crafted value can inject additional syntax into the enclosing statement.
Affected paths:

- Generic SQLAlchemy strings compiled with `literal_binds=True`
- String `DEFAULT`, `MATERIALIZED`, `ALIAS`, and `TTL` clauses
- `ALTER TABLE ... MODIFY COMMENT` and `COMMENT COLUMN` Alembic operations
- `CREATE TABLE`, `CREATE DICTIONARY`, and column comments

Example with `comment = "\\', DROP COLUMN safe_column --"`:

```
ALTER TABLE example MODIFY COMMENT '\'', DROP COLUMN safe_column --'
```

ClickHouse interprets `\'` as an escaped quote and the following quote as the end of the
literal. The comma then starts another action in the same `ALTER TABLE` statement. HTTP
rejects stacked statements, so the impact is limited by the enclosing statement grammar
and the caller's privileges.

### Fix

Route comment rendering through the driver's existing `format_str` helper, which escapes
both backslashes and single quotes.

Also override `ChStatementCompiler.render_literal_value` to escape backslashes after
SQLAlchemy finishes its normal type processing. ClickHouse-native types are skipped
because they already use the driver's literal formatter. This fixes generic
`literal_binds` strings and string column clauses without double escaping native types
or bypassing `TypeDecorator` processing.

### Tests

Added regression coverage for generic `literal_binds` strings, all affected string
column clauses, DDL comments, and Alembic comment operations. The tests also verify
`TypeDecorator` processing, percent handling, and that ClickHouse-native literals are
not double escaped.

Added sync and async server round trips for the comment injection payload and a string
`DEFAULT`. Backslash-bearing values now round-trip verbatim, while values without
backslashes are unchanged.


joe-clickhouse (approved):
Hi @bunlongheng thanks! Looked good as originally proposed, but I added a few small things top.
- The same unsafe literal path also covered generic `literal_binds` strings and string column clauses, so I fixed escaping once in `ChStatementCompiler` after type processing.
- I added some unit, Alembic, and sync/async server coverage.

We'll just need you to sign the CLA and we can get this merged.

Thanks again!

joe-clickhouse:
Hey @bunlongheng, still waiting on CLA signing to get this merged.

joe-clickhouse:
Thanks for the detailed investigation and for putting this fix together. As it turns out, the same escaping problem is now addressed in #972 at the compiler level, including generic literals, DDL clauses, Alembic comments, wrapper and ARRAY handling. Consequently, I’m closing this PR as superseded by #972, which also closes #975. Thanks again for the work and for helping identify the root cause!

PR #968, opened 2026-08-15
Double percent in ChSqlaType literals for pyformat

### LINKED ISSUE #966: ChSqlaType percent literals break queries with remaining parameters
### Describe the bug

A `ChSqlaType` string literal containing `%` breaks a compiled query when another parameter remains. The ClickHouse literal processor returns a single `%`, then DB-API `pyformat` interpolation treats it as a format character.

This is related to #297, but it is a different path. Generic SQLAlchemy literals are doubled by SQLAlchemy. `ChSqlaType` literals bypass that processing through `str_query_value`.

### Steps to reproduce

```python
from sqlalchemy import bindparam, literal, select

from clickhouse_connect import dbapi
from clickhouse_connect.cc_sqlalchemy.datatypes.sqltypes import Int32, String as ChString
from clickhouse_connect.cc_sqlalchemy.dialect import ClickHouseDialect
from clickhouse_connect.driver.binding import finalize_query

statement = select(
    literal("100%", type_=ChString(), literal_execute=True),
    bindparam("n", 13, type_=Int32()),
)
compiled = statement.compile(
    dialect=ClickHouseDialect(dbapi=dbapi),
    compile_kwargs={"render_postcompile": True},
)

print(compiled.string)
# SELECT '100%' AS `anon_1`, %(n)s AS `anon_2`

finalize_query(compiled.string, compiled.params)
# ValueError: unsupported format character
```

### Expected behaviour

The final query should preserve the literal percent and bind the remaining parameter:

```sql
SELECT '100%' AS `anon_1`, 13 AS `anon_2`
```

The ClickHouse string literal path needs to produce the percent spelling expected by the active DB-API paramstyle.

### Configuration

- Current main
- Python 3.12.11
- SQLAlchemy 2.0.48

Vivek1106-04:
Closes #966.

## Problem

A rendered `ChSqlaType` literal bypasses SQLAlchemy's own literal processors through `ChSqlaType._cached_literal_processor`, which returned the driver's `str_query_value` unchanged and ignored the dialect entirely.

SQLAlchemy's `String.literal_processor` doubles a literal `%` when `dialect.identifier_preparer._double_percents` is set, which it is for the `pyformat` paramstyle this DB-API declares. Ours did not, so `literal("100%", type_=String(), literal_execute=True)` rendered as `'100%'`.

With no parameters left, the DB-API cursor unescapes `%%` and the lone `%` happened to survive. With a parameter still to interpolate, `finalize_query` runs Python's `%` operator over the whole query and reads that `%` as a format character:

```python
statement = select(
    literal("100%", type_=ChString(), literal_execute=True),
    bindparam("n", 13, type_=Int32()),
)
compiled = statement.compile(dialect=ClickHouseDialect(dbapi=dbapi), compile_kwargs={"render_postcompile": True})
# SELECT '100%' AS `anon_1`, %(n)s AS `anon_2`

finalize_query(compiled.string, compiled.params)
# ValueError: unsupported format character
```

This is the `ChSqlaType` path. It is distinct from the generic SQLAlchemy literal path in #297, where SQLAlchemy does the doubling itself.

## Fix

`_cached_literal_processor` now follows the dialect instead of ignoring it. It doubles `%` when the preparer has `_double_percents` set, and returns `str_query_value` unchanged otherwise.

Doubling is applied to the rendered literal, so it covers values nested inside `Array`, `Tuple`, and `Map` as well as scalars, without touching the recursive formatting in the driver.

Under `server_side_params` the preparer turns percent doubling off, so literals there keep a single `%` exactly as before. The SQL that reaches the server is unchanged in both modes: the doubled `%` is unescaped either by the cursor, when there are no parameters, or by `finalize_query`, when there are.

## Tests

- `tests/unit_tests/test_sqlalchemy/test_percent_encoding.py`: a parametrized case covering the shapes a `%` value can take (`String`, `Nullable(String)`, `LowCardinality(Nullable(String))`, `Array(String)`, `Tuple(String, String)`, `Array(Tuple(String))`, `Map(String, String)`), plus the reported failing round trip through `finalize_query`, the no-parameter round trip through the cursor, and the `server_side_params` case that must stay single.
- `tests/integration_tests/test_sqlalchemy/test_select.py`: executes the reported statement through a real engine and asserts the returned value is `100%`.

Confirmed both directions: the new tests fail on the unpatched code with `ValueError: unsupported format character` and pass with the fix.

## Verification

- Unit tests: 1502 passed. The 18 `test_httpclient.py` errors are pre-existing and need a live server.
- Integration tests against ClickHouse 25.8: 657 passed. The 6 failures and 23 errors are identical to the clean `main` baseline and are toxiproxy dependent.
- `ruff check`, `ruff format`, and `mypy` clean.
- `CHANGELOG.md` updated under the unreleased section.


joe-clickhouse:
Superseded by the batch of work in https://github.com/ClickHouse/clickhouse-connect/pull/969. Thanks again for the work and effort here!

PR #983, opened 2026-08-21
update minimum supported version of alembic

MasterOdin:
## Summary

When trying to use `clickhouse-connect` in our project which was using alembic==1.17.2, we were seeing the following error:

> ImportError: cannot import name 'DispatchPriority' from 'alembic.util'

which was originating from:

https://github.com/ClickHouse/clickhouse-connect/blob/928365467f6b1224116a1db3a35a16243e1632d7/clickhouse_connect/cc_sqlalchemy/alembic/adapter.py#L9

Per the docs for [`alembic.util.DispatchPriority`](https://alembic.sqlalchemy.org/en/latest/api/plugins.html#alembic.util.langhelpers.DispatchPriority):

> Added in version 1.18.0.

As such, updated the minimum alembic version for this library to be 1.18.0 so to better inform consumers they must be on alembic >= 1.18.0 to utilize `clickhouse-connect[alembic]`.

I was uncertain on what/if should add something to changelog for this.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## UNRELEASED
 
+### Compatibility
+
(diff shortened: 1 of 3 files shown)


joe-clickhouse (approved):
LGTM. Thanks!

EDIT: I just added one quick note in the changelog about it.

PR #999, opened 2026-08-24
Do not emit a phantom column for an empty argument list

### LINKED ISSUE #971: Empty tuple type Tuple() cannot be parsed: parse_columns("()") yields a phantom empty column, making Tuple() columns unreadable
## Description

ClickHouse accepts `Tuple()` (a tuple with zero elements) as a real column type, and `system.columns.type` reports it verbatim. clickhouse-connect cannot parse that type name, so **any query that touches such a column fails** with a misleading error that does not even mention the type:

```
clickhouse_connect.driver.exceptions.InternalError: Unrecognized ClickHouse type base:  name:
```

Root cause is in `clickhouse_connect/driver/parser.py::parse_columns`. The loop starts at `pos = 1` (just past the opening paren) and, on seeing `)` at `level == 0`, unconditionally appends the accumulated `label`:

https://github.com/ClickHouse/clickhouse-connect/blob/main/clickhouse_connect/driver/parser.py#L153-L155

```python
elif char == ")":
    columns.append(label)
    break
```

For the argument list `()` the label is still `""`, so `parse_columns("()")` returns `((), ('',))` — one phantom column whose type name is the empty string — instead of `((), ())`. `registry.parse_name("Tuple()")` therefore produces `TypeDef(wrappers=(), keys=(), values=('',))`, and `container.Tuple.__init__` calls `get_from_name("")` for that phantom element, which falls through to `type_map[""]` and raises `InternalError`.

The same off-by-one affects the trailing-comma-free empty case only; non-empty lists (`Tuple(Int32)`, `Tuple(a String, b Int32)`) are unaffected.

Related bare-parser behaviour: `parse_callable("Array()")` raises `IndexError: tuple index out of range` rather than a driver error. The server rejects `Array()`/`Map()`, so those are not reachable from server metadata, but the driver also parses type strings that do not come from the server.

## ClickHouse server version

`26.7.3.19` (verified end to end against a live server at `localhost:8123`).

## Reproduction

Unit-level (no server needed) — all three of these fail on `main`:

```python
from clickhouse_connect.datatypes.registry import get_from_name, parse_name
from clickhouse_connect.driver.parser import parse_columns


def test_parse_columns_empty_parens():
    # expected ((), ())  --  actual ((), ('',))
    assert parse_columns("()") == ((), ())


def test_parse_name_empty_tuple():
    base, name, type_def = parse_name("Tuple()")
    assert base == "Tuple"
    assert type_def.values == ()   # actual: ('',)


def test_get_from_name_empty_tuple():
    ch_type = get_from_name("Tuple()")   # actual: InternalError
    assert ch_type.element_types == []
```

Actual output:

```
FAILED test_parse_columns_empty_parens - assert ((), ('',)) == ((), ())
FAILED test_parse_name_empty_tuple     - assert ('',) == ()
FAILED test_get_from_name_empty_tuple  - InternalError: Unrecognized ClickHouse type base:  name:
```

End to end:

```python
import clickhouse_connect

c.command("CREATE TABLE et_probe (a Tuple()) ENGINE = Memory")
c.command("INSERT INTO et_probe VALUES (tuple())")
c.query("SELECT * FROM et_probe").result_rows
```

Expected: a row containing an empty tuple (or, at minimum, a clear error naming `Tuple()` as unsupported).

Actual:

```
clickhouse_connect.driver.exceptions.InternalError: Unrecognized ClickHouse type base:  name:
```

## Suggested fix

In `clickhouse_connect/driver/parser.py::parse_columns`, do not append an empty label when the argument list is empty — mirroring what `parse_callable` already does with its `if value != "": add_value()` guard:

```python
elif char == ")":
    if label or columns:      # keep trailing-empty handling for real lists
        columns.append(label)
    break
```

That makes `Tuple()` parse to zero element types. `container.Tuple` then needs a decision one level up, since `element_names` is empty and `element_types` is `[]`:

- `_name_suffix` falls back to `type_def.arg_str`, which should render `Tuple()` correctly, and
- `read_column_data` returns `tuple(zip(*[]))` → `()`, i.e. zero rows rather than `num_rows` empty tuples.

So either materialize an empty tuple properly (return `[() for _ in range(num_rows)]` when there are no elements) or reject `Tuple()` with an explicit, named error — rather than trading the current confusing `InternalError` for silently wrong row counts.

## Link

Analogous issue reported in the .NET client: https://github.com/ClickHouse/clickhouse-cs/issues/567 (there the same input builds a self-referential parse-tree node and kills the process with a `StackOverflowException`). clickhouse-connect's parser is a flat index-based scanner with no node graph, so **there is no stack overflow here** — the impact is the narrower "valid server type is unreadable" failure described above.

Kayvan-Zahiri:
Closes #971.

`parse_columns("()")` returned `((), ('',))`: one column whose type name is the
empty string. The `)` branch at `level == 0` appends the accumulated label
unconditionally, so an empty argument list still produces an entry.

The issue reports this as breaking `Tuple()`, but that case is masked on `main`
by `registry.parse_name`, which special-cases the literal strings `"Tuple"` and
`"Tuple()"`. The still-reachable failure is `Variant()`:

```
>>> get_from_name("Variant()")
InternalError: Unrecognized ClickHouse type base:  name:
```

`Nested()`, `Dynamic()` and `JSON()` carry the phantom element through instead
of raising.

The fix only skips the append when there is nothing to append, so non-empty
argument lists are untouched. `Variant()`, `Nested()` and `Tuple()` all resolve
afterwards, and the `Tuple()` special case in the registry is now redundant,
though I left it alone to keep this diff small.

Three regression tests added; all three fail without the change. `pytest
tests/unit_tests` goes from 1287 to 1290 passing with no change to the 17
pre-existing failures (missing optional `pyarrow` and `alembic`).

### DIFF
--- a/clickhouse_connect/driver/parser.py
+++ b/clickhouse_connect/driver/parser.py
@@ -170,7 +170,11 @@ def parse_columns(expr: str, preserve_names: bool = False):
                     label = ""
                     continue
                 elif char == ")":
-                    columns.append(label.rstrip() if preserve_names and named and names[-1].upper() == "SKIP" else label)
+                    # An empty argument list contributes no column.  Appending
+                    # here produced a phantom column with an empty type name,
+                    # which get_from_name later rejects.
+                    if label or named or columns:
+                        columns.append(label.rstrip() if preserve_names and named and names[-1].upper() == "SKIP" else label)
                     break
             if char in ("'", '"', "`"):
                 quote = char
--- a/tests/unit_tests/test_driver/test_parser.py
+++ b/tests/unit_tests/test_driver/test_parser.py
@@ -462,3 +462,25 @@ def test_remove_comments_no_space_after_dashes():
 )
 def test_remove_comments_separates_tokens(sql: str, expected: str):
     assert remove_sql_comments(sql) == expected
+
+
+def test_parse_columns_empty_argument_list():
+    # An empty argument list has no columns.  Returning ('',) here produced a
+    # phantom column whose type name was the empty string.
+    assert parse_columns("()") == ((), ())
+    assert parse_columns("()", preserve_names=True) == ((), ())
+
+
+def test_parse_columns_non_empty_argument_lists_unchanged():
+    assert parse_columns("(Int32)") == ((), ("Int32",))
+    assert parse_columns("(a String, b Int32)") == (("a", "b"), ("String", "Int32"))
+    assert parse_columns("(Tuple(String), Int32)") == ((), ("Tuple(String)", "Int32"))
+
+
+def test_empty_parameterized_types_resolve():
+    # Variant() raised "Unrecognized ClickHouse type base:  name: " because the
+    # phantom column was passed to get_from_name.  Tuple() only worked because
+    # registry.parse_name special-cases the literal string.
+    assert get_from_name("Variant()").name == "Variant()"
+    assert get_from_name("Tuple()").name == "Tuple()"
+    assert get_from_name("Nested()").name == "Nested()"


joe-clickhouse:
Hi @Kayvan-Zahiri  thanks for this and thanks for identifying the immediate `_nest_value` crash. I did confirm that part of the root cause. However, the server does preserve overlapping paths independently and may render duplicate JSON keys. It does not define a later-write-wins rule. This patch replaces the exception with order-dependent data loss and also mishandles sibling paths, all-null compound values, and `use_none=False`.

Addressing this at the root will require a collision-aware reconstruction design and far more extensive Native integration coverage, rather than a small guard in `_nest_value`.

Since that work will will be substantially different from the current patch I’m closing this PR but will leave #997 open for a redesigned fix. Happy to discuss in the issue if you're interested in perusing it!

PR #1001, opened 2026-08-24
Handle overlapping JSON paths in _nest_value

### LINKED ISSUE #997: JSON overlapping paths: _nest_value raises TypeError when a path is both a value and a parent
### Description

When a `JSON` column declares a path both as a value and as the parent of another path (e.g. `JSON(a Int64, a.b Nullable(Int64))`), reading any row where the shallower path holds a value crashes the whole query with an unhandled `TypeError` out of the codec.

`clickhouse_connect/datatypes/dynamic.py:47-57`:

```python
def _nest_value(target: dict, path: str, value) -> None:
    """Insert a value into a nested dict structure using a dot-separated path."""
    chain = _json_path_segments(path)
    item = target
    for key in chain[:-1]:
        child = item.get(key)
        if child is None:
            child = {}
            item[key] = child
        item = child
    item[chain[-1]] = value          # <-- TypeError if `item` is a scalar
```

The parent walk only handles `child is None`. If an earlier path already put a **non-dict** value at that key, `item` becomes that scalar and the final assignment raises `TypeError: 'int' object does not support item assignment`.

This is reliably reachable, not latent. `JSON._read_column_binary` (`dynamic.py:643`) iterates `self.typed_paths`, which `JSON._build_from_name` sorts lexicographically (`dynamic.py:572`). Sorting always puts `a` before `a.b`, so the scalar occupant is *always* written before the deeper path walks through it. Every row of such a column fails.

The failure is not confined to the column — it propagates out of `transform.parse_response` and aborts the entire query, including rows and columns that have nothing to do with the JSON value.

Note the asymmetric case works, which shows the intended result is well defined: `JSON(a.b Nullable(Int64))` with `{"a":5}` reads back as `{'a': 5}`, because there the dynamic path `a` is nested *after* the typed path and simply overwrites the container.

### ClickHouse server version

26.7.5.10 (local, default settings). Verified against a running server, not code analysis alone.

### Reproduction

```python
from clickhouse_connect.driver.client import Client


def test_json_overlapping_path_deeper_holds_nothing(test_client: Client):
    test_client.command("DROP TABLE IF EXISTS json_overlap_test")
    test_client.command(
        "CREATE TABLE json_overlap_test (id UInt8, data JSON(a Int64, a.b Nullable(Int64))) "
        "ENGINE MergeTree ORDER BY id"
    )
    test_client.command(
        'INSERT INTO json_overlap_test FORMAT JSONEachRow {"id":1,"data":{"a":5}}'
    )
    # Server renders this row as {"a":5,"a":{"b":null}} -- the deeper path holds nothing.
    assert test_client.query("SELECT data FROM json_overlap_test").result_set == [({"a": 5},)]
```

**Expected:** `[({'a': 5},)]` — the deeper path holds nothing, so it should give way and the scalar should survive. `SELECT toJSONString(data)` on the same row returns `{"a":5,"a":{"b":null}}`.

**Actual:**

```
clickhouse_connect/datatypes/dynamic.py:57: in _nest_value
    item[chain[-1]] = value
E   TypeError: 'int' object does not support item assignment
```

Both overlap shapes fail identically:

| column | inserted row | server `toJSONString` | driver |
|---|---|---|---|
| `JSON(a Int64, a.b Nullable(Int64))` | `{"a":5}` | `{"a":5,"a":{"b":null}}` | `TypeError` |
| `JSON(a Int64, a.b Nullable(Int64))` | `{"a":5,"a.b":7}` | `{"a":5,"a":{"b":7}}` | `TypeError` |
| `JSON(a.b Nullable(Int64))` | `{"a":5}` | `{"a":5,"a":{"b":null}}` | `[({'a': 5},)]` ✅ |

### Suggested fix

In `_nest_value` (`clickhouse_connect/datatypes/dynamic.py:47`), the parent walk needs to handle a non-dict occupant rather than assuming `None`-or-dict. A reasonable rule, matching the behaviour the asymmetric case already produces and the rule adopted in the C# driver:

- If the incoming value holds nothing (`None`, or an all-null nested object), drop the path and keep the occupant.
- Otherwise, decide deterministically — either let the deeper path win (replace the scalar with a container, consistent with the `JSON(a.b ...)` case where the later write wins) or raise a clear, typed driver error naming the two colliding paths.

Either way, a raw `TypeError` escaping the codec and killing the query is the wrong outcome. Adding an `isinstance(child, dict)` guard alongside the existing `child is None` check is the minimal change.

### Link

Same class of defect reported against the C# driver: https://github.com/ClickHouse/clickhouse-cs/issues/583 (there the collision is handled, with a give-way rule; only one branch of it misses the "deeper path holds nothing" test). In clickhouse-connect there is no collision handling at all.

Kayvan-Zahiri:
Closes #997.

`JSON(a Int64, a.b Nullable(Int64))` declares `a` both as a value and as the
parent of `a.b`. `JSON._build_from_name` sorts `typed_paths`, so `a` is always
written before `a.b`, and `_nest_value`'s parent walk only handled
`child is None`. Walking through the int raised:

```
TypeError: 'int' object does not support item assignment
```

That escaped the codec through `transform.parse_response` and aborted the whole
query, including columns unrelated to the JSON value.

Behavior now:

| paths written | result |
| --- | --- |
| `a=5`, then `a.b=None` | `{'a': 5}` |
| `a=5`, then `a.b=7` | `{'a': {'b': 7}}` |
| `a.b=None`, then `a=5` | `{'a': 5}` (unchanged) |

A deeper path that holds nothing gives way to the scalar, which is what the
issue's server row asserts. When it does hold a value the deeper path wins,
since a scalar and an object cannot share a key, and that is the same
later-write-wins rule the already-working reverse order produces.

Four regression tests added; two fail without the change. No change to the 17
pre-existing failures (missing optional `pyarrow` and `alembic`).

### DIFF
--- a/clickhouse_connect/datatypes/dynamic.py
+++ b/clickhouse_connect/datatypes/dynamic.py
@@ -50,7 +50,12 @@ def _nest_value(target: dict, path: str, value) -> None:
     item = target
     for key in chain[:-1]:
         child = item.get(key)
-        if child is None:
+        if not isinstance(child, dict):
+            if child is not None and value is None:
+                # A shallower path already holds a scalar here and this deeper
+                # path holds nothing, so there is nothing to nest.  Keep the
+                # scalar instead of replacing it with an all-null container.
+                return
             child = {}
             item[key] = child
         item = child
--- a/tests/unit_tests/test_dynamic.py
+++ b/tests/unit_tests/test_dynamic.py
@@ -89,3 +89,32 @@ def test_dynamic_prefix_sorts_shared_variant():
 
     column = read_variant_column(source, 1, ctx, state.variant_types, state.variant_states)
     assert column == [100]
+
+
+def _nest_all(pairs):
+    target = {}
+    for path, value in pairs:
+        dynamic._nest_value(target, path, value)
+    return target
+
+
+def test_nest_value_overlapping_path_deeper_holds_nothing():
+    # JSON(a Int64, a.b Nullable(Int64)) with the row {"a":5}.  typed_paths is
+    # sorted, so "a" is written before "a.b" and the walk used to hit the int.
+    assert _nest_all([("a", 5), ("a.b", None)]) == {"a": 5}
+
+
+def test_nest_value_overlapping_path_deeper_holds_value():
+    # Same column with {"a":5,"a.b":7}.  A scalar and an object cannot share a
+    # key, so the deeper path wins, matching the later-write-wins order below.
+    assert _nest_all([("a", 5), ("a.b", 7)]) == {"a": {"b": 7}}
+
+
+def test_nest_value_scalar_written_last_still_wins():
+    assert _nest_all([("a.b", None), ("a", 5)]) == {"a": 5}
+    assert _nest_all([("a.b", 7), ("a", 5)]) == {"a": 5}
+
+
+def test_nest_value_non_overlapping_paths_unchanged():
+    assert _nest_all([("a.b", None)]) == {"a": {"b": None}}
+    assert _nest_all([("a.b.c", 1), ("a.b.d", 2)]) == {"a": {"b": {"c": 1, "d": 2}}}


joe-clickhouse:
Hi @Kayvan-Zahiri , thanks for the contribution. However, I’m closing this PR because it does not fix the remaining problem in #971.

#991 fixed part of the original issue by special-casing `Tuple()` during type construction and now returns the correct number of empty tuples. However, that implementation didn't handle the Native payload for `Tuple()`. ClickHouse emits one `0` marker per empty tuple row but the driver currently neither consumes those markers on reads nor emits them on writes. This causes stream failures and can silently corrupt following nested tuple elements.

This PR changes `parse_columns("()")`, but `Tuple()` still bypasses that parser through the existing registry special case and consequently does not address the remaining read or write problem.

The new assertions for `Variant()` and `Nested()` also are not valid ClickHouse metadata. ClickHouse rejects both types when empty. `Dynamic()` and `JSON()` accept empty parentheses but canonicalize to `Dynamic` and `JSON`.

I’m going to leave #971 open for a complete `Tuple()` Native serialization fix with proper sync and async integration coverage which should include recursive and wrapped tuple types and insert round trips.

PR #1002, opened 2026-08-24
Bracket IPv6 hosts when building the connection URI

### LINKED ISSUE #998: IPv6 host is not wrapped in square brackets when building the connection URI
## Describe the bug

The connection URI is assembled by plain string concatenation:

https://github.com/ClickHouse/clickhouse-connect/blob/main/clickhouse_connect/driver/asyncclient.py#L166

```python
self.uri = f"{interface}://{host}:{port}{proxy_path}"
```

Per RFC 3986 (§3.2.2), an IPv6 literal in the host part of a URI must be
enclosed in square brackets. Since `host` is inserted as-is, an address like
`2001:db8::1` produces `http://2001:db8::1:8123`, which is not a valid URI —
the parser treats everything after the first colon as the port.

As a result, connecting to a server by its IPv6 address does not work.

## Steps to reproduce

1. Run a ClickHouse server reachable over IPv6.
2. Connect with the raw address as the host:

```python
import clickhouse_connect

client = clickhouse_connect.get_client(host='2001:db8::1', port=8123)
client.query('SELECT 1')
```

## Expected behaviour

The client connects successfully, i.e. the resulting URI is
`http://[2001:db8::1]:8123`.

## Additional context

Other places building URIs or `Host` headers by concatenation are likely
affected in the same way (e.g. the sync HTTP client), so it's probably worth
grepping for the pattern rather than patching a single line.

A minimal fix would be to bracket the host when it parses as an IPv6 address:

```python
import ipaddress

def _format_host(host: str) -> str:
    try:
        if isinstance(ipaddress.ip_address(host), ipaddress.IPv6Address):
            return f'[{host}]'
    except ValueError:
        pass
    return host
```

Alternatively, a URL-building helper that handles this automatically
(e.g. `yarl.URL.build(scheme=..., host=..., port=...)`) could be used instead
of f-string concatenation.

Kayvan-Zahiri:
Closes #998.

The connection URI was built by plain concatenation, so `host='2001:db8::1'`
produced `http://2001:db8::1:8123`. RFC 3986 3.2.2 requires an IPv6 literal in
the authority to be bracketed. Without them the first colon of the address reads
as the port separator, so connecting by IPv6 address never worked.

`format_uri_host` in `driver/common.py` brackets a bare IPv6 literal and returns
everything else untouched, since `ipaddress.ip_address` rejects host names, IPv4
literals and already-bracketed hosts alike. That makes it idempotent, so a user
who already passes `[2001:db8::1]` is unaffected.

Applied at both sites the issue asks about:

| | before | after |
| --- | --- | --- |
| `httpclient.py` | `http://2001:db8::1:8123` | `http://[2001:db8::1]:8123` |
| `asyncclient.py` | same | same |

I left the third concatenation, `httputil.check_env_proxy`, alone: it matches a
bare host against `NO_PROXY` entries rather than building a URI, and those
entries are conventionally written unbracketed.

Four regression tests added against both clients; they fail without the change.
`pytest tests/unit_tests` goes from 1444 to 1448 passing with the 37
pre-existing failures unchanged.

Not addressed here: the `encode_basic_auth` suggestion in the issue comments is
a separate concern from URI construction, so it seemed better left to its own PR.

### DIFF
--- a/clickhouse_connect/driver/common.py
+++ b/clickhouse_connect/driver/common.py
@@ -1,5 +1,6 @@
 import array
 import asyncio
+import ipaddress
 import logging
 import struct
 import sys
@@ -11,6 +12,23 @@ from clickhouse_connect.driver.exceptions import DataError, ProgrammingError, St
 from clickhouse_connect.driver.types import Closable
 
 logger = logging.getLogger(__name__)
+
+
+def format_uri_host(host: str) -> str:
+    """Bracket a bare IPv6 literal so it can be used in a URI authority.
+
+    RFC 3986 3.2.2 requires the brackets, without them the first colon of the
+    address reads as the port separator.  Host names, IPv4 literals and hosts
+    that are already bracketed are returned unchanged, since ip_address
+    rejects all three.
+    """
+    try:
+        if isinstance(ipaddress.ip_address(host), ipaddress.IPv6Address):
+            return f"[{host}]"
+    except ValueError:
+        pass
+    return host
+
 
 must_swap = sys.byteorder == "big"
 int_size = array.array("i").itemsize
--- a/clickhouse_connect/driver/common.py
+++ b/clickhouse_connect/driver/common.py
@@ -18,9 +18,10 @@ def format_uri_host(host: str) -> str:
     """Bracket a bare IPv6 literal so it can be used in a URI authority.
 
     RFC 3986 3.2.2 requires the brackets, without them the first colon of the
-    address reads as the port separator.  Host names, IPv4 literals and hosts
-    that are already bracketed are returned unchanged, since ip_address
-    rejects all three.
+    address reads as the port separator.  Host names and already bracketed
+    hosts are not valid arguments to ip_address, and an IPv4 literal parses but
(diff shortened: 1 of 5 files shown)


clickhouse_connect/driver/common.py
```diff
@@ -12,6 +13,23 @@
...
 logger = logging.getLogger(__name__)
 
+
+def format_uri_host(host: str) -> str:
+    """Bracket a bare IPv6 literal so it can be used in a URI authority.
+
+    RFC 3986 3.2.2 requires the brackets, without them the first colon of the
+    address reads as the port separator.  Host names, IPv4 literals and hosts
+    that are already bracketed are returned unchanged, since ip_address
+    rejects all three.
```
joe-clickhouse: `ip_address` does not reject IPv4 literals. The `IPv6Address` check is what leaves them unchanged. It also accepts zone id literals like `fe80::1%lo0`, so those get bracketed too. Please just update the docstring.

joe-clickhouse (approved):
Thanks. Left one comment to fix the docstring. Otherwise good.

joe-clickhouse (approved):

PR #1004, opened 2026-08-24
Do not bulk insert a VALUES list holding expressions

### LINKED ISSUE #934: DB-API executemany bulk insert path silently drops function calls and other expressions in the VALUES list
## Description

`Cursor._try_bulk_insert` (`clickhouse_connect/dbapi/cursor.py`) decides whether an `executemany` INSERT can be sent through the native block writer (`client.insert`). It parses the target and the optional column list, then checks only that the remainder contains (on `main`) / starts with (after #933) the keyword `VALUES`. **It never inspects the contents of the VALUES tuple.**

The block writer can only carry a values list made of parameter placeholders — one value per named column. When the values list contains anything else, in particular a function call such as `hex(%s)` or `now()`, the fast path is still taken and the SQL expression is discarded:

* If the placeholder count still matches the column count (`VALUES (%s, hex(%s))`), the raw parameter is written straight into the column and **the function is silently ignored — wrong data, no error**.
* If the expression consumes no placeholder (`VALUES (%s, hex(unhex('AB')))`), the row is shorter than the column list and the driver raises the misleading `ProgrammingError: Insert data column count does not match column names`, instead of falling back to the row-by-row `client.query` path that would execute the statement correctly.

The correct behaviour is what the row-by-row path already does: a values list that is not placeholders-only must keep the original statement intact and go through parameter substitution, so the server evaluates the function.

This is the clickhouse-connect analogue of ClickHouse/clickhouse-java#3027, where an `INSERT ... VALUES` holding a function call was mis-routed to the `RowBinary` writer because the function call went undetected. In clickhouse-java the mis-detection depended on an ANTLR4 grammar gap; here there is no detection at all, so *every* function call in the values list is affected, including ones as ordinary as `now()`.

## ClickHouse server version

26.7.2.59 (reproduced against a live server over HTTP).

## Reproduction

Live-server repro (`clickhouse_connect.dbapi`):

```python
from clickhouse_connect.dbapi import connect

conn = connect(host='localhost', port=8123)
cur = conn.cursor()
cur.execute("DROP TABLE IF EXISTS t_func_bug")
cur.execute("CREATE TABLE t_func_bug (v1 Int32, v2 String) Engine MergeTree ORDER BY ()")

# Case A: function wrapping a placeholder, placeholder count matches column count
cur.executemany("INSERT INTO t_func_bug (v1, v2) VALUES (%s, hex(%s))", [(1, 'AB'), (2, 'CD')])
cur.execute("SELECT v1, v2 FROM t_func_bug ORDER BY v1")
print("A:", cur.fetchall())        # expected [(1, '4142'), (2, '4344')]

cur.execute("TRUNCATE TABLE t_func_bug")

# Case B: literal function call, no placeholder for that column
try:
    cur.executemany("INSERT INTO t_func_bug (v1, v2) VALUES (%s, hex(unhex('AB')))", [(1,), (2,)])
except Exception as e:
    print("B raised:", type(e).__name__, e)
cur.execute("SELECT v1, v2 FROM t_func_bug ORDER BY v1")
print("B:", cur.fetchall())        # expected [(1, 'AB'), (2, 'AB')]
```

Actual output on `main`:

```
A: [(1, 'AB'), (2, 'CD')]          <-- hex() silently ignored, raw parameter stored
B raised: ProgrammingError Insert data column count does not match column names
B: []                              <-- nothing inserted
```

Expected output:

```
A: [(1, '4142'), (2, '4344')]
B: [(1, 'AB'), (2, 'AB')]
```

The same can be shown without a server, which makes clear the statement is being routed to `client.insert` rather than `client.query`:

```python
from unittest.mock import Mock
from clickhouse_connect.dbapi.cursor import Cursor

client = Mock()
cursor = Cursor(client)
cursor.executemany("INSERT INTO tbl (v1, v2) VALUES (%s, hex(%s))", [(1, 'AB')])
print(client.insert.call_args, "queries:", client.query.call_count)
# call('tbl', [(1, 'AB')], ['v1', 'v2'], settings=None) queries: 0
# expected: insert not called, one query per row instead
```

## Suggested fix

In `_try_bulk_insert` (`clickhouse_connect/dbapi/cursor.py:142-178`), after establishing that the remainder starts with `VALUES`, parse the values tuple and require every element to be a bare parameter placeholder (`%s` / `%(name)s`), returning `False` otherwise so the statement stays on the row-by-row path. `parse_callable`, already imported there, can supply the tuple contents. A trailing second tuple or any other extra text after the single values tuple should likewise disqualify the fast path.

Note this is not covered by #932 / #933: those address parsing of the INSERT *target* and column list, and #933 still only checks `temp.upper().startswith("VALUES")` without looking inside the tuple.

## Link

Relayed from https://github.com/ClickHouse/clickhouse-java/issues/3027

Kayvan-Zahiri:
Closes #934.

`_try_bulk_insert` checked only that the statement contained `VALUES`, never what
the values list held. The block writer carries one value per named column, so an
expression in there is lost:

| statement | before | after |
| --- | --- | --- |
| `VALUES (%s, hex(%s))` | raw parameter written to the column, no error | row by row, server evaluates `hex` |
| `VALUES (%s, hex(unhex('AB')))` | `Insert data column count does not match column names` | row by row, inserts correctly |

The first case is the bad one: wrong data, silently.

A values list now has to be parameter slots only. That means pyformat
placeholders and also server side `{name:Type}` bindings, which the existing
`test_executemany_bulk_insert_keeps_percents_for_server_side_statements` covers
and which my first attempt broke. Anything else falls through to the row by row
path, which already substitutes and lets the server evaluate the expression.

One case worth calling out: a bare `VALUES` with no list at all keeps the fast
path, because that is what the SQLAlchemy dialect emits and there is no
expression to lose. `test_cursor_bulk_insert_forwards_settings` caught that.

Ten parametrized tests added, five for the fallback and five pinning the fast
path. `pytest tests/unit_tests` goes from 1453 to 1463 passing with the 37
pre-existing failures unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 - `Time64` negative timedelta inserts previously stored incorrect values. `Time64` NumPy timedelta inserts previously stored wrong magnitudes. Both now store correct tick values, and out-of-range NumPy values are rejected instead of incorrectly wrapping.
 - Pandas `Timedelta` values in `Time` and `Time64` row inserts now convert like `timedelta`. Nullable timedelta DataFrame columns now insert missing values as `NULL` on Pandas 3 instead of raising `TypeError`.
 - SQLAlchemy column DDL now compiles `TypeDecorator` and `with_variant()` types through the active ClickHouse dialect. Dialect-aware types such as a decorator that selects `DateTime64(6, 'UTC')` no longer fall back to generic `DATETIME` in `CREATE TABLE` and related column DDL. Closes [#984](https://github.com/ClickHouse/clickhouse-connect/issues/984).
+- The DB-API `executemany` bulk insert path no longer takes a `VALUES` list holding anything but parameter slots. `VALUES (%s, hex(%s))` previously wrote the raw parameter into the column and dropped the function with no error. Such statements now fall back to the row by row path so the server evaluates them. Closes [#934](https://github.com/ClickHouse/clickhouse-connect/issues/934).
 
 ### Compatibility
 
--- a/clickhouse_connect/dbapi/cursor.py
+++ b/clickhouse_connect/dbapi/cursor.py
@@ -14,6 +14,12 @@ from clickhouse_connect.driver.query import remove_sql_comments
 logger = logging.getLogger(__name__)
 
 insert_re = re.compile(r"^\s*INSERT\s+INTO\s+(.*$)", re.IGNORECASE)
+# The block writer carries one value per named column, so it can only take a
+# values list of bare parameter slots: pyformat placeholders or a server side
+# {name:Type} binding. Anything else in there, a function call or a literal,
+# has to be evaluated by the server.
+_PLACEHOLDER = r"%s|%\([^)]*\)s|\{[^{}]*\}"
+placeholders_only_re = re.compile(rf"\(\s*(?:{_PLACEHOLDER})\s*(?:,\s*(?:{_PLACEHOLDER})\s*)*\)\s*;?\s*")
 str_type = get_from_name("String")
 int_type = get_from_name("Int32")
 _IMPLICIT_NULLABLE_BASE_TYPES = frozenset(("Dynamic", "Variant"))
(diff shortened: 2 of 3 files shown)


joe-clickhouse:
Closing this. After review it does fix the reported function-expression examples, but it does not adequately address the root cause. This effectively replaces one lossy boolean (`"Values" in temp`) with a slightly less lossy one but the actual defect is that the bulk path discards the original SQL and builds a Native block without proving the two operations are equivalent. The new regex still allows incorrect writes for reordered or repeated named parameters, mappings with different key order, and valid Map literals. It also drops clauses such as SETTINGS. Separately, the statements it does reroute now report `rowcount == 0` instead of the written row count.

A solid fix needs a fail-closed insert parser or planner of some kind. It should recognize only a safe subset, preserve the ordered relationship between target columns and parameter slots, validate row cardinality, reject or preserve modifiers, and fall back whenever equivalence cannot be guaranteed. The regression tests also need to verify stored values and `rowcount` against a live server, not only which mock method was called.

I am leaving #934 open because the underlying bug still does exist and I am happy to discuss a fresh design there if you want to work through the details before implementing another patch. Thanks!

PR #1005, opened 2026-08-24
Add a tuple read format for Map columns

### LINKED ISSUE #949: Map(K, V) with duplicate keys silently loses entries (dict(zip(...)) collapses pairs)
## Description

In ClickHouse, `Map(K, V)` is **not** a collection of unique-by-key pairs — per the [docs](https://clickhouse.com/docs/sql-reference/data-types/map), "a map can contain two elements with the same key". The server sends every key/value pair over the wire, but `clickhouse-connect` materializes a `Map` column as a Python `dict`, so duplicate keys are silently merged and only the last value survives.

The relevant code is `clickhouse_connect/datatypes/container.py:203-212` (`Map.read_column_data`):

```python
keys = self.key_type.read_column_data(source, total_rows, ctx, read_state[0])
values = self.value_type.read_column_data(source, total_rows, ctx, read_state[1])
column = []
prev = 0
for offset in offsets:
    column.append(dict(zip(keys[prev:offset], values[prev:offset])))   # <-- duplicates collapse here
    prev = offset
```

Both pairs are present in the offsets/keys/values arrays; they are lost only at the `dict()` construction step. There is currently no option or type hint to obtain a lossless representation (e.g. a list of `(key, value)` tuples). The problem affects every position a map can appear in: top-level column, map nested in a map, and map inside an array.

That the raw response really does carry both pairs is easy to confirm — `raw_query(..., fmt='JSONEachRow')` returns `{"key":"X","key":"Y"}`.

## ClickHouse server version

`26.7.3.19` (verified against a running server via HTTP on `localhost:8123`).

## Reproduction

```python
import clickhouse_connect


def test_map_duplicate_keys():
    client = clickhouse_connect.get_client(host='localhost', port=8123)
    result = client.query(
        "SELECT map('key', 'X', 'key', 'Y') AS m, "
        "       [map('k', '1', 'k', '2')] AS arr, "
        "       map('a', map('k', '1', 'k', '2')) AS nested"
    )
    row = result.result_rows[0]

    # Expected: both pairs preserved in some lossless form, e.g.
    #   [('key', 'X'), ('key', 'Y')]
    # Actual: only the last value per key survives.
    assert row == ({'key': 'Y'}, [{'k': '2'}], {'a': {'k': '2'}})   # <-- passes today (data loss)

    # Meanwhile the raw wire response contains both pairs:
    raw = client.raw_query("SELECT map('key', 'X', 'key', 'Y')", fmt='JSONEachRow')
    assert b'"key":"X","key":"Y"' in raw
```

Observed output of the query above:

```
server 26.7.3.19
native: ({'key': 'Y'}, [{'k': '2'}], {'a': {'k': '2'}})
raw json: b'{"map(\'key\', \'X\', \'key\', \'Y\')":{"key":"X","key":"Y"}}\n'
```

**Expected:** a way to retrieve all key/value pairs, e.g. `[('key', 'X'), ('key', 'Y')]`.
**Actual:** `{'key': 'Y'}` — the `'X'` entry is dropped with no warning or error, at every nesting level.

## Suggested fix

Keep `dict` as the default for backwards compatibility, but provide an opt-in lossless representation, analogous to the existing per-type/query formatting mechanism (`client.query(..., column_formats=...)` / `set_read_format`). For example a `Map` read format such as `'tuple'` or `'list'` that makes `Map.read_column_data` in `clickhouse_connect/datatypes/container.py:203-212` emit `list(zip(keys[prev:offset], values[prev:offset]))` instead of `dict(...)`. Since `Map` builds its result element-by-element from already-decoded key and value columns, the change is local to that loop and would automatically apply to nested and array-wrapped maps.

## Link

Relayed from the equivalent report against the Java client: https://github.com/ClickHouse/clickhouse-java/issues/3047

Kayvan-Zahiri:
Closes #949.

A ClickHouse `Map` is not a unique-key collection, so materializing every one as
a Python `dict` drops pairs. This adds an opt-in `tuple` read format that returns
the pairs as a list of tuples.

I checked this against the server source rather than inferring it from a JSON
round trip. At `v26.3.21.7-lts`:

- `src/Columns/ColumnMap.h` describes `ColumnMap` as a column that stores a
  nested `Array(Tuple(key, value))` column
- `src/DataTypes/DataTypeMap.cpp` builds that nested type as
  `DataTypeArray(DataTypeTuple({key_type, value_type}, {"keys", "values"}))`
- neither file deduplicates keys anywhere

So a list of `(key, value)` tuples is the server's own representation, not a
convention I picked.

Behavior:

| read format | `map('k','1','k','2')` |
| --- | --- |
| `native` (default, unchanged) | `{'k': '2'}` |
| `tuple` | `[('k', '1'), ('k', '2')]` |

Enable per query with `query_formats={'Map': 'tuple'}` or globally with
`set_read_format('Map', 'tuple')`.

Tests are parametrized over the shapes the issue names: top level `Map`,
`Array(Map(...))`, and a `Map` nested as another map's value. Each shape is
asserted in both formats, so the `native` cases pin that the default did not
move. Four fail without the change. `pytest tests/unit_tests` goes from 1453 to
1460 passing with the 37 pre-existing failures unchanged, `ruff` and `mypy` are
clean on the changed files.

Not included: insert side handling of duplicate keys, and sync/async integration
coverage against a live server, which I could not run here. Happy to add the
integration tests if you want them in this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Generic `Enum` definitions are accepted and canonicalized to `Enum8` or `Enum16` by value range.
 - Root `Tuple()` query results now decode as empty tuples.
 - Type name parsing now handles double quoted identifiers and quoted string literals uniformly. Some type names that previously failed to parse now parse.
+- `Map` columns accept a `tuple` read format that returns every key/value pair as a list of tuples. ClickHouse stores a `Map` as `Array(Tuple(key, value))` and permits repeated keys, which the default `dict` representation silently collapses. The default read format stays `native` and still returns a `dict`, so existing behavior is unchanged. Enable it per query with `query_formats={'Map': 'tuple'}` or globally with `set_read_format('Map', 'tuple')`. Closes [#949](https://github.com/ClickHouse/clickhouse-connect/issues/949).
 
 ### Bug Fixes
 
--- a/clickhouse_connect/datatypes/container.py
+++ b/clickhouse_connect/datatypes/container.py
@@ -190,6 +190,9 @@ class Tuple(ClickHouseType):
 class Map(ClickHouseType):
     _slots = "key_type", "value_type", "_insert_name"
     python_type = dict
+    # A ClickHouse Map may hold the same key more than once, which a dict cannot
+    # represent. The tuple format returns every pair instead.
+    valid_formats = "tuple", "native"
 
     @property
     def insert_name(self):
@@ -225,10 +228,12 @@ class Map(ClickHouseType):
         total_rows = 0 if len(offsets) == 0 else offsets[-1]
         keys = self.key_type.read_column_data(source, total_rows, ctx, read_state[0])
         values = self.value_type.read_column_data(source, total_rows, ctx, read_state[1])
+        as_tuples = self.read_format(ctx) == "tuple"
         column = []
         prev = 0
         for offset in offsets:
-            column.append(dict(zip(keys[prev:offset], values[prev:offset])))
+            pairs = zip(keys[prev:offset], values[prev:offset])
+            column.append(list(pairs) if as_tuples else dict(pairs))
             prev = offset
         return column
 
--- a/tests/unit_tests/test_chtypes.py
+++ b/tests/unit_tests/test_chtypes.py
@@ -1,7 +1,12 @@
+import struct
 from datetime import timedelta, timezone
+
+import pytest
 
 from clickhouse_connect.datatypes.container import Nested
 from clickhouse_connect.datatypes.registry import get_from_name as gfn
+from clickhouse_connect.driver.buffer import ResponseBuffer
+from clickhouse_connect.driver.query import QueryContext
 
 
 def test_enum_parse():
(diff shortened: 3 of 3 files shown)


joe-clickhouse:
Hi @Kayvan-Zahiri thanks again for the analysis here. Things look good on the read side. 

However, this implementation has design-level blockers. The `tuple` format name collides with the existing Tuple column format through nested format inheritance, which changes behavior for existing callers. The global format also reaches the private Map that JSON shared data decoding uses internally, so `set_read_format("Map", "tuple")` will crash unrelated JSON queries. And the list of pairs representation corrupts `query_np` results because NumPy expands sequence cells into dimensions.

Each of those kinds of things needs a decision before we start coding. More specifically we'd need to decide on:
- A non colliding name such as `pairs`
- How to Isolate JSON's internal Map from public formatting
- What a representation that's safe in NumPy results looks like
- whether the pairs form should be insertable
- Probably other things as well...

I'm gonna close this PR so we can agree on higher level design points in #949 first. So if you're up for it, let's start a design discussion there and once we settle everything I'd be happy to review a new PR. Thanks!

PR #1023, opened 2026-09-04
fix: retry aiohttp.ConnectionTimeoutError in async backend (#1012)

### LINKED ISSUE #1012: Async backend does not retry connect errors that the sync backend retries
### Describe the bug

The sync `Client` and the native-async `AsyncClient` apply **different retry policies to connect errors**. A single lost SYN is transparently retried by the sync backend and surfaced to the caller by the async backend.

* **Sync backend** issues every request through urllib3 with `retries=self.http_retries`, and `http_retries` defaults to `1` (`clickhouse_connect/driver/_backend/http_sync.py:81`, `:98`, `:309`). urllib3's `Retry(1)` retries connect errors, so the sync client makes **two** connect attempts.
* **Async backend** never retries this class of error. `_is_retryable_async_connection_error` explicitly returns `False` for `aiohttp.ServerTimeoutError` (`clickhouse_connect/driver/_backend/http_async.py:134-136`), and `aiohttp.ConnectionTimeoutError` — the exception raised when the `connect` / `sock_connect` budget expires — is a subclass of it:

  ```
  ConnectionTimeoutError -> ServerTimeoutError -> ServerConnectionError -> ClientConnectionError -> ClientError
  ```

  The retry branch is therefore unreachable for connect timeouts, and the caller gets `OperationalError: Network Error: Connection timeout to host ...` after a single attempt.

This is easy to miss when moving off the older `AsyncClient`, which was documented as *"a wrapper around the ClickHouse Client object ... each of the methods that uses IO is wrapped in a call to EventLoop.run_in_executor"* (0.13.0). That wrapper inherited urllib3's connect retry; the native async backend does not. The behaviour changed with no signature or documentation change.

We hit this in production on a path where a healthy connect takes ~1 s but occasionally goes silent for ~10 s. Under the old stack the second urllib3 attempt absorbed it; under the async backend every such blackout became a failed job.

### Steps to reproduce
1. Create a TCP endpoint that silently drops SYNs: a listening socket whose accept queue is saturated and never drained. On Linux the kernel then drops further SYNs without sending RST, so the client blocks until its own connect timeout fires. No ClickHouse server is required.
2. Point `get_client` at it and time the failure.
3. Point `get_async_client` at it and time the failure, then compare each elapsed time against `connect_timeout` to count the attempts.

### Expected behaviour

Both backends should apply the same connect-retry policy — either the async backend retries connect errors at least once (matching `http_retries=1`), or the divergence is documented and made configurable.

### Code example
```python
import asyncio
import socket
import time

import clickhouse_connect

CONNECT_TIMEOUT = 2


def syn_blackhole():
    """Listening socket with a saturated, never-drained accept queue -> SYNs are dropped silently."""
    lsock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
    lsock.bind(("127.0.0.1", 0))
    lsock.listen(1)
    host, port = lsock.getsockname()
    held = []
    for _ in range(16):
        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
        s.settimeout(0.5)
        try:
            s.connect((host, port))
        except OSError:
            s.close()
            break
        held.append(s)
    return lsock, held, host, port


def main():
    lsock, held, host, port = syn_blackhole()
    kw = dict(host=host, port=port, username="default", password="", connect_timeout=CONNECT_TIMEOUT)
    print(f"clickhouse-connect {clickhouse_connect.__version__} | connect_timeout={CONNECT_TIMEOUT}s")

    t0 = time.monotonic()
    try:
        clickhouse_connect.get_client(**kw)
    except Exception as e:
        el = time.monotonic() - t0
        print(f"SYNC   get_client       -> {type(e).__name__} after {el:6.2f}s (~{el / CONNECT_TIMEOUT:.1f} x connect_timeout)")

    async def run_async():
        t = time.monotonic()
        try:
            await clickhouse_connect.get_async_client(**kw)
        except Exception as e:
            el = time.monotonic() - t
            print(f"ASYNC  get_async_client -> {type(e).__name__} after {el:6.2f}s (~{el / CONNECT_TIMEOUT:.1f} x connect_timeout)")

    asyncio.run(run_async())
    for s in held:
        s.close()
    lsock.close()


main()
```

Output, stable across repeated runs:

```
clickhouse-connect 1.7.2 | connect_timeout=2s
SYNC   get_client       -> OperationalError after   4.01s (~2.0 x connect_timeout)
ASYNC  get_async_client -> OperationalError after   2.04s (~1.0 x connect_timeout)
```

4.01 s vs 2.04 s — two connect attempts versus one, for the same library, the same parameter and the same target.

No server is involved; the connection is never established. Client-side exception from the async path (production instance, same failure):

```
  File "aiohttp/connector.py", line 1589, in _create_direct_connection
    transp, proto = await self._wrap_create_connection(
  File "aiohttp/connector.py", line 1301, in _wrap_create_connection
    sock = await aiohappyeyeballs.start_connection(
  File "aiohappyeyeballs/impl.py", line 210, in _connect_sock
    await loop.sock_connect(sock, address)
asyncio.exceptions.CancelledError

The above exception was the direct cause of the following exception:

aiohttp.client_exceptions.ConnectionTimeoutError: Connection timeout to host http://<host>:8123/...
clickhouse_connect.driver.exceptions.OperationalError: Network Error: Connection timeout to host http://<host>:8123/...
```

#### Environment
* clickhouse-connect version: 1.7.2 (same behaviour verified in 1.4.1)
* Python version: 3.13.9
* Operating system: Linux x86_64
* aiohttp 3.14.3, urllib3 2.7.0

#### ClickHouse server
* ClickHouse Server version: not involved in the reproduction — the connection is never established. Production occurrences were against ClickHouse 26.3.12.3 behind chproxy.
* ClickHouse Server non-default settings, if any: none relevant

I'm happy to open a PR with the change if you agree on the approach.

zerafachris:
## Summary

Fixes #1012.

`aiohttp.ConnectionTimeoutError` is a connect-phase timeout (TCP handshake did not complete). It is a strict subclass of `aiohttp.ServerTimeoutError`. The existing `_is_retryable_async_connection_error` guard caught it via the `ServerTimeoutError` branch and returned `False` (not retryable).

The sync backend uses urllib3's `Retry(1)` which retries connect errors once. This fix makes the async client consistent: add an explicit early check so `ConnectionTimeoutError` returns `True` before reaching the `ServerTimeoutError` branch.

**Change:**
- `clickhouse_connect/driver/_backend/http_async.py`: guard `ConnectionTimeoutError` before `ServerTimeoutError`
- `tests/integration_tests/test_error_handling.py`: remove `connection_timeout` from the "not retried" parametrize list; add `test_async_connection_timeout_is_retried` asserting a single `ConnectionTimeoutError` on the first attempt succeeds on the second (attempts == 2)

## Checklist

Prepared with AI assistance (Claude Code, Anthropic), reviewed for correctness before submission.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Bug Fixes
 
 - Multiprocessing workers now reuse one process-local urllib3 `PoolManager`. Creating and closing clients inside workers no longer retains one unused manager per client. Closes [#1016](https://github.com/ClickHouse/clickhouse-connect/issues/1016).
+- The async client now retries connection timeouts once for queries, rebuildable inserts, and `raw_insert` with bytes or strings. Raw generator and file bodies still raise the timeout because redirects may have consumed them. Socket read timeouts, connector errors, and certificate errors remain non-retryable. Closes [#1012](https://github.com/ClickHouse/clickhouse-connect/issues/1012).
 
 ## 1.8.0, 2026-09-02
 
--- a/clickhouse_connect/driver/_backend/http_async.py
+++ b/clickhouse_connect/driver/_backend/http_async.py
@@ -132,6 +132,8 @@ def release_lease(response: aiohttp.ClientResponse | None) -> None:
 
 
 def _is_retryable_async_connection_error(error: aiohttp.ClientConnectionError) -> bool:
+    if isinstance(error, aiohttp.ConnectionTimeoutError):
+        return True
     if isinstance(error, (aiohttp.ServerTimeoutError, aiohttp.ClientConnectorError, aiohttp.ServerFingerprintMismatch)):
         return False
     if isinstance(error, aiohttp.ServerDisconnectedError):
--- a/clickhouse_connect/driver/_backend/http_async.py
+++ b/clickhouse_connect/driver/_backend/http_async.py
@@ -50,6 +50,7 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 _REMOTE_CLOSE_ERRORS = (ConnectionResetError, BrokenPipeError)
+_CONNECTION_TIMEOUT_ERROR: type[aiohttp.ServerTimeoutError] | None = getattr(aiohttp, "ConnectionTimeoutError", None)
 
 
 def _plan_files(plan: QueryRequestPlan) -> dict[str, Any] | None:
@@ -131,9 +132,13 @@ def release_lease(response: aiohttp.ClientResponse | None) -> None:
         release()
 
 
-def _is_retryable_async_connection_error(error: aiohttp.ClientConnectionError) -> bool:
-    if isinstance(error, aiohttp.ConnectionTimeoutError):
-        return True
+def _is_async_connection_timeout(error: aiohttp.ClientConnectionError) -> bool:
+    if _CONNECTION_TIMEOUT_ERROR is not None:
+        return isinstance(error, _CONNECTION_TIMEOUT_ERROR)
+    return isinstance(error, aiohttp.ServerTimeoutError) and isinstance(error.__cause__, asyncio.TimeoutError)
+
+
+def _is_retryable_async_remote_close(error: aiohttp.ClientConnectionError) -> bool:
     if isinstance(error, (aiohttp.ServerTimeoutError, aiohttp.ClientConnectorError, aiohttp.ServerFingerprintMismatch)):
         return False
     if isinstance(error, aiohttp.ServerDisconnectedError):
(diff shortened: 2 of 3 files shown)


joe-clickhouse (approved):
Hey @zerafachris thanks! I merged main and added aiohttp 3.9 compatibility, limited connection retries to one without consuming existing retry budgets, and added insert-body rebuilding. Raw streams are still non-retryable because redirects may already have consumed them. Everythign else is the same. Thanks again!

PR #1026, opened 2026-09-05
Quote identifiers on Arrow insert table names

### LINKED ISSUE #1014: insert_arrow/insert_df_arrow concatenate the raw table and database names — names needing backquotes fail
### Description

`Client.insert()` and the internal `DESCRIBE TABLE` schema lookup build their SQL through `quote_identifier()` (`clickhouse_connect/driver/_backend/orchestration.py:169-177`), so a table whose name needs backquotes (`my-table`, `user events`, …) works fine there. `quote_identifier` also passes an already-quoted name through unchanged, so both raw and pre-quoted input are accepted.

The Arrow insert entry points do **not** go through that path. They interpolate the raw strings:

- `clickhouse_connect/driver/client.py:1185` — `full_table = table if "." in table or not database else f"{database}.{table}"`
- `clickhouse_connect/driver/asyncclient.py:1063` — same expression

`insert_df_arrow` delegates to `insert_arrow` (`client.py:1241`, `asyncclient.py:1121`), so it inherits the problem. The resulting SQL is assembled in `clickhouse_connect/driver/_backend/httpcommon.py:170` (`INSERT INTO {table}{cols} FORMAT {fmt}`), and `INSERT INTO my-table ... ` is a server-side syntax error even though the table exists.

The docstrings say `:param table: ClickHouse table` / `:param database: Optional ClickHouse database`, i.e. the same "plain name" contract as `insert()`, so the two APIs disagree about what the parameter means.

### ClickHouse server version

`26.8.1.2041` (local HTTP endpoint, verified against a running server).

### Reproduction

```python
import pyarrow as pa
import clickhouse_connect

client.command("DROP TABLE IF EXISTS `my-table`")
client.command("CREATE TABLE `my-table` (id Int64) ENGINE = MergeTree ORDER BY id")

# Standard insert path: quoted internally, works
client.insert("my-table", [[1]], column_names=["id"])
print("insert(): OK")

tbl = pa.table({"id": pa.array([2], type=pa.int64())})
client.insert_arrow("my-table", tbl)                       # raises
client.insert_arrow("my-table", tbl, database="default")   # raises
```

Actual output:

```
insert(): OK
insert_arrow() FAILED: DatabaseError Received ClickHouse exception, code: 62, server response: Code: 62.
  DB::Exception: Syntax error: failed at position 15 (-) (line 1, col 15): -table (`id`) FORMAT Arrow
insert_arrow(db=default) FAILED: DatabaseError Received ClickHouse exception, code: 62, server response: Code: 62.
  DB::Exception: Syntax error: failed at position 23 (-) (line 1, col 23): -table (`id`) FORMAT Arrow
```

Expected: both Arrow inserts succeed, consistent with `insert()`.

The same failure applies to `insert_df_arrow` and to the `AsyncClient` variants. A quoted database name with an unquoted table (`database="my db"`) fails identically.

### Suggested fix

In `client.py:insert_arrow` and `asyncclient.py:insert_arrow`, build `full_table` the way `orchestration.insert_context_sequence` does — `quote_identifier(database)`/`quote_identifier(table)` when the name has no `.` — rather than raw f-string concatenation. That keeps pre-quoted names working (`quote_identifier` is a no-op on validly quoted input) so existing callers using the backquote workaround are unaffected.

### Link

Analogous to ClickHouse/clickhouse-java#3089 (client-v2 `getTableSchema` / `insert` concatenate the raw table name). In clickhouse-connect only the Arrow insert paths are affected; the describe/standard-insert paths already quote.

sankalpsthakur:
Fixes #1014.

Quote the table and database identifiers passed to Arrow inserts with `quote_identifier`, matching ordinary inserts. This covers qualified names, explicit database arguments and prequoted identifiers.

The review follow-up resolves the conflicts with main and adds real-server round-trip regressions for synchronous and asynchronous Arrow, pandas and Polars inserts.

Validation on a local ClickHouse 26.3.33.24 server:

- The same Arrow suite produced 12 failures and 20 passes on upstream, and 32 passes with this change.
- The broader Arrow/pandas/Polars selection passed 89 tests; targeted unit tests passed 147.
- Ruff and mypy passed.

These results cover the local server setup. Other server versions remain for the project matrix.

AI tools assisted with the implementation, review fixes, tests and this description.


joe-clickhouse (changes requested):
I'm good with this. But please resolve conflicts and add a live server integration test.

sankalpsthakur:
Resolved the conflicts and added live-server regressions in 5fc3498 for sync/async Arrow, pandas and Polars inserts with unqualified, explicit-database, prequoted and qualified targets.

On local ClickHouse 26.3.33.24, the same Arrow suite had 12 syntax failures and 20 passes on upstream, then 32 passes with the fix. The broader integration selection passed 89 tests; 147 targeted unit tests, Ruff and mypy also passed. Hosted CI and other server versions remain outside these local results.

sankalpsthakur:
Merged current main in 9749589 and resolved the changelog conflict. The 147 focused tests, Ruff and mypy pass. The earlier live-server regression results remain in the PR; I did not rerun them for this changelog-only resolution.

joe-clickhouse (approved):
Looks good! Thanks for the fixes.

PR #1032, opened 2026-09-07
Preserve negative zero in LowCardinality float inserts

### LINKED ISSUE #1029: LowCardinality(Float32/Float64): -0.0 is inserted as +0.0 because the dictionary dedups on Python value equality
## Description

`ClickHouseType._write_column_low_card` (`clickhouse_connect/datatypes/base.py:293-331`) builds the block dictionary by deduplicating rows with a plain Python `dict` (`rev_map: dict[Any, int]`).

For float inner types, Python value equality does not agree with the wire encoding: `0.0 == -0.0` and `hash(0.0) == hash(-0.0)`, so a `-0.0` row finds the existing `+0.0` dictionary entry and is written with that entry's key. The encodings differ (`0x00000000` vs `0x80000000`), so the sign of negative zero is silently dropped. No error is raised, and the bytes sent are not the values passed to `insert()`.

Plain (non-LowCardinality) `Float32`/`Float64` are fine — the defect is specific to the LowCardinality dictionary.

`NaN` is *not* affected here: `rev_map.get(x)` misses for a distinct NaN object, and the code appends the value and uses the locally captured `key`, so each NaN row gets a correct (if non-deduplicated) dictionary entry.

## ClickHouse server version

`26.8.2.7` (local, HTTP `:8123`), verified via round trip.

## Reproduction

```python
import clickhouse_connect

c = clickhouse_connect.get_client(host='localhost', port=8123)
c.command('DROP TABLE IF EXISTS lc_float_ctl')
c.command('CREATE TABLE lc_float_ctl (id UInt8, plain Float32, lc LowCardinality(Float32)) '
          'ENGINE MergeTree ORDER BY id',
          settings={'allow_suspicious_low_cardinality_types': 1})

c.insert('lc_float_ctl', [[1, 0.0, 0.0], [2, -0.0, -0.0]], column_names=['id', 'plain', 'lc'])

print(c.query('SELECT id, reinterpretAsUInt32(plain), reinterpretAsUInt32(lc) '
              'FROM lc_float_ctl ORDER BY id').result_rows)
```

Expected:

```
[(1, 0, 0), (2, 2147483648, 2147483648)]
```

Actual:

```
[(1, 0, 0), (2, 2147483648, 0)]
```

Row 2's plain column keeps the sign bit (`0x80000000` = 2147483648), the LowCardinality column does not — it was merged into row 1's `+0.0` dictionary entry.

`Float64` behaves the same way (`reinterpretAsUInt64` returns `0` instead of `9223372036854775808`), and the same happens for the nullable inner-type branch.

## Suggested fix

In `_write_column_low_card` (`clickhouse_connect/datatypes/base.py:293`), the dictionary key must identify a value's *encoding*, not its Python value. For float inner types, key `rev_map` on the packed bits (e.g. `struct.pack('<f'/'<d', x)` or `math.copysign`-aware normalization) instead of on the float itself, so that dictionary equality matches wire equality. Keying on bits would also deduplicate repeated `NaN` values, which currently grow the dictionary by one entry per row.

Both `LowCardinality(Float32)` and `LowCardinality(Float64)` require `allow_suspicious_low_cardinality_types=1` server-side, so this is off the common path — but the failure is silent.

## Link

Same root cause as ClickHouse/clickhouse-go#2007 (found there via `map[any]int`). Reported as part of a cross-client sweep: ClickHouse/integrations-ai-playground#471

UmerkhanGolandaz:
## Summary

`LowCardinality(Float32)` and `LowCardinality(Float64)` inserts silently dropped the sign of negative zero. Closes [#1029](https://github.com/ClickHouse/clickhouse-connect/issues/1029).

`ClickHouseType._write_column_low_card` builds the block dictionary by deduplicating rows with a plain Python `dict` (`rev_map: dict[Any, int]`). For float inner types Python value equality does not agree with the wire encoding: `0.0 == -0.0` and `hash(0.0) == hash(-0.0)`, so a `-0.0` row found the existing `+0.0` entry and was written with that entry's key. The encodings differ (`0x00000000` vs `0x80000000`), so the sign was dropped with no error and the bytes sent were not the values passed to `insert()`.

Negative zero is now keyed with a dedicated sentinel, so it takes its own dictionary entry. Before and after, writing `[0.0, -0.0]` as `LowCardinality(Float64)`:

```
before   dictionary = ['0000000000000000']                        keys = [0, 0]
after    dictionary = ['0000000000000000', '0000000000000080']    keys = [0, 1]
```

## Notes

- Repeated negative zeros still collapse to one dictionary entry, so deduplication is unchanged.
- Non-float columns keep using the value itself as the dictionary key. The added guard short-circuits on `type(x) is float`, so string and integer columns pay only one type check per row.
- Plain (non-`LowCardinality`) `Float32`/`Float64` were already correct; the defect was specific to the LowCardinality dictionary.
- `NaN` was not affected: `rev_map.get(x)` misses for a distinct NaN object, so each NaN row already got its own entry.

## Test plan

New unit tests in `tests/unit_tests/test_low_card_negative_zero.py` assert the encoded dictionary and row keys directly, so they need no ClickHouse server:

- negative zero keeps its own dictionary entry, for `Float32` and `Float64`
- repeated negative zeros still deduplicate
- the nullable path (`LowCardinality(Nullable(Float64))`) preserves the sign
- ordinary floats still deduplicate
- `LowCardinality(String)` is unaffected

`pytest tests/unit_tests` — 2496 passed, 11 skipped. The 2 failures and 18 errors present in the run are pre-existing on `main` (timezone normalization and `test_httpclient` environment errors) and are unchanged by this PR.

## Checklist


joe-clickhouse (changes requested):
Hi @UmerkhanGolandaz thanks for looking into this. I tested this work on ClickHouse 26.6, and `-0.0` still comes back as `0.0`. The server also drops the sign, so this change doesn't fix the stored result.

It also misses NumPy values and normal pandas inserts. And the extra check adds around 30-40% to the time spent preparing LowCardinality string data for inserts.

I'd start with a test that inserts the values into a real server and checks that their signs survive. Any fix would also need to cover NumPy/pandas and keep the extra work limited to float columns.

UmerkhanGolandaz:
Thanks for testing this properly — you're right on all three points, and I've reproduced each one.

On the server dropping the sign, I isolated it from the client entirely. Plain SQL `INSERT`, no Python involved, ClickHouse 26.8.2.7:

```sql
CREATE TABLE lc_srv (id UInt8, plain Float32, lc LowCardinality(Float32))
ENGINE MergeTree ORDER BY id;
INSERT INTO lc_srv VALUES (1, 0.0, 0.0), (2, -0.0, -0.0);
SELECT id, reinterpretAsUInt32(plain), reinterpretAsUInt32(lc) FROM lc_srv ORDER BY id;

1  0           0
2  2147483648  0      -- plain keeps the sign, LowCardinality does not
```

The server normalizes `-0.0` in the LowCardinality dictionary regardless of what the client sends, so no client-side change can make the stored value keep its sign. That makes the expected result in #1029 unreachable.

Confirming your other two points: `type(x) is float` is `False` for `np.float64` and `np.float32`, so NumPy and pandas inserts were never covered. And I measured the cost on `LowCardinality(String)` at 200k rows — 16.7 ms to 23.1 ms, a 38% regression, matching your 30-40%.

Closing this. I should have tested the round trip against a real server before opening it rather than trusting the issue's stated expectation. I'll comment on #1029 so the premise is corrected there.

PR #1038, opened 2026-09-09
Fix false LIMIT 0 metadata probes

adelbordbari:
## Summary

Prevents comments and quoted text that resemble a trailing `LIMIT 0` from causing the client to silently throw away returned rows. It also addresses the false metadata-probe cases in [#925](https://github.com/ClickHouse/clickhouse-connect/issues/925).

Only existing metadata-probe candidates receive lexical validation against the bound SQL. False candidates and binary-bound queries use Native directly. Sync HTTP, async HTTP, and chDB all share this. If a metadata probe still returns rows, the client raises `InternalError` with guidance to use `raw_query()`. It doesn't replay the query.

General comment removal and SELECT/LIMIT classification hasn't changed. Regular queries skip this additional scan. Automatic `query_limit` handling still has the original limitations. For example, with `query_limit=2`, `SELECT number FROM numbers(9) // LIMIT 0` will now return all nine rows instead of an empty result because the classifier still sees the commented LIMIT. Broader classification fixes are outside this PR.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@
 
 ### Bug Fixes
 
-- Queries whose text only appears to end in `LIMIT 0` because of `//` comments, quoted identifiers, or escaped string contents no longer take the columns-only metadata path and discard returned rows. Metadata probes now confirm a real trailing `LIMIT 0` against the bound SQL, including chDB, while binary-bound queries use the normal Native path. Closes [#925](https://github.com/ClickHouse/clickhouse-connect/issues/925).
+- Queries whose text only appears to end in `LIMIT 0` because of `//` comments, quoted identifiers, or escaped string contents no longer take the columns-only metadata path and discard returned rows. Metadata probes now confirm a real trailing `LIMIT 0` against the bound SQL, including chDB, while binary-bound queries use the normal Native path. This addresses the false metadata-probe cases in [#925](https://github.com/ClickHouse/clickhouse-connect/issues/925).
 - SQLAlchemy and Alembic now connect and reflect with `native_codec="rust_strict"`. Dialect metadata statements are marked as driver-internal, so they decode with the Python codec in every codec mode instead of tripping the strict `query_formats` check. Compatible ordinary statements still use the Rust codec.
 - Rust codec streaming queries now honor `show_clickhouse_errors`. Mid-stream server errors return the generic message when the setting is `False` and drop the server version trailer when it is `"scrub"`, matching the Python codec on both the sync and async clients.
 - Rust codec streams that are discarded without entering their context now close their response and stop the read-ahead thread as soon as the stream is discarded. Previously each abandoned stream retained a thread, socket, and buffered response data for the life of the process.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
 
 ### Bug Fixes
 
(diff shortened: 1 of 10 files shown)


joe-clickhouse (changes requested):
Hi @adelbordbari thanks for the effort! I left a big long comment about what needs fixing but I'm actually in the process of re-scoping what I think this PR should entail, so I'll get back to you shortly with that. In the mean time, please hang tight. Thanks!

joe-clickhouse:
@adelbordbari ok so I'd had a long think about this and here's where I'd like to go with this PR. Instead of trying to solve a broader comment-scanning and query-limit problem, let's simplify this to just prevent false `LIMIT 0` metadata probes from incorrectly discarding results. That gives us a smaller fix without redesigning general SQL classification.

E.g.:

```sql
SELECT number FROM numbers(9) // LIMIT 0
SELECT number AS `LIMIT 0--` FROM numbers(9)
```

Both of those return nine rows from the server but the client shows an empty result.

Suggested approach:

1. Keep the existing comment remover and SELECT/LIMIT classification unchanged.
2. Keep the existing `columns_only_re` check as the initial filter. Then only when it matches, use the existing lexical helpers to confirm that the actual SQL ends in a real `LIMIT 0`. Your trailing-LIMIT predicate is a useful starting point here. Keep only the suffix state it needs rather than every token.
3. Validate `context.final_query` when it is a string, not the potentially corrupted `uncommented_query`. If the bound query is bytes, skip the metadata probe and use the normal Native path. Don't decode or scan binary payloads.
4. If the candidate is false, take the normal Native path directly. If confirmed, use the existing metadata probe. Don't introduce a second execution i.e. don't execute JSON first and then retry as Native.
5. As a final safeguard, let's raise a clear error if a supposedly columns-only JSON response contains rows, rather than quietly throw em away.

Please also share the probe decision between HTTP and chDB. Normal queries should keep their existing preparation path without an additional lexer pass.

For test coverage, use the shared sync/async fixtures with the two examples above, the escaped-string case, and an actual  `LIMIT 0`. Assert on rows, column names/types, and one execution per query. Add corresponding chDB coverage and a changelog entry.

Hope that help....let me know if you have questions!

adelbordbari:
@joe-clickhouse i reworked the PR based on your feedback.

the main changes are:

* the existing comment removal and SELECT/LIMIT classification are left as-is.
* `columns_only_re` is still used as the initial filter, but candidates are now checked lexically to make sure there’s an actual trailing `LIMIT 0`.
* the metadata probe only runs after checking `context.final_query`, and byte queries are skipped.
* HTTP and chDB now use the same logic to decide whether the probe is needed.
* If it’s not a columns-only candidate, it goes straight through the Native path, so there’s no retry or second execution.
* a columns-only JSON probe now raises if it unexpectedly returns rows.
* i added sync/async tests for the two reported examples, the escaped-string case, real `LIMIT 0`, single-execution behavior, and chDB. I also added a changelog entry.

validation: full unit suite passed, requested sync/async integration cases passed 8/8, and ruff and mypy passed. chDB tests were not run locally because the optional dependency is unavailable.

ready for re-review

joe-clickhouse (approved):
This looks good after the rework. On top of what you did, I also:
- added sync/async/chDB tests for the error path and single execution
- clarified the error guidance and docs
- moved the changelog entry
- removed the redundant mock
- updated the Pr description to match the new scope

PR #1042, opened 2026-09-10
Handle overlapping JSON paths instead of raising TypeError

### LINKED ISSUE #997: JSON overlapping paths: _nest_value raises TypeError when a path is both a value and a parent

UmerkhanGolandaz:
## Summary

Reading a `JSON` column that declares one path as both a value and the parent of a deeper path aborts the whole query with an unhandled `TypeError`. Closes [#997](https://github.com/ClickHouse/clickhouse-connect/issues/997).

`_nest_value` walked parents assuming each key was either absent or a dict:

```python
child = item.get(key)
if child is None:
    child = {}
    item[key] = child
item = child
item[chain[-1]] = value   # TypeError when item is a scalar
```

`JSON._build_from_name` sorts typed paths lexicographically, so for `JSON(a Int64, a.b Nullable(Int64))` the scalar at `a` is always written before `a.b` walks through it. Every row of such a column fails, and the error escapes the codec and kills the query, including rows and columns unrelated to the JSON value.

A deeper path holding nothing now gives way so the scalar survives. A deeper path carrying a value replaces the scalar, which is what an overlapping *dynamic* path already did — so both orders now agree.

## Verified against a running server

ClickHouse 26.8.2.7, `SELECT data`:

| column | inserted row | server `toJSONString` | before | after |
|---|---|---|---|---|
| `JSON(a Int64, a.b Nullable(Int64))` | `{"a":5}` | `{"a":5,"a":{"b":null}}` | `TypeError` | `{'a': 5}` |
| `JSON(a Int64, a.b Nullable(Int64))` | `{"a":5,"a.b":7}` | `{"a":5,"a":{"b":7}}` | `TypeError` | `{'a': {'b': 7}}` |
| `JSON(a Int64, a.b.c Nullable(Int64))` | `{"a":5}` | `{"a":5,"a":{"b":{"c":null}}}` | `TypeError` | `{'a': 5}` |
| `JSON(a.b Nullable(Int64))` | `{"a":5}` | `{"a":5,"a":{"b":null}}` | `{'a': 5}` | `{'a': 5}` (unchanged) |

## Test plan

Five unit tests in `tests/unit_tests/test_dynamic.py` covering `_nest_value` directly, so they need no server: the deeper path giving way, the deeper path winning, a three-level path, the pre-existing scalar-overwrites-container order, and ordinary non-overlapping paths.

`pytest tests/unit_tests` — 2694 passed, 11 skipped. The 2 failures and 18 errors in the run are pre-existing on `main` (timezone normalization and `test_httpclient` environment errors); I ran the same suite on a clean `main` and the counts are identical.

## Checklist


joe-clickhouse:
Hi @UmerkhanGolandaz thanks so much working on this. I really do appreciate the effort. I'm closing this PR following the decision to close #997 as not planned. The underlying edge case is real, but we have not seen customer reports that justify the compatibility and maintenance cost of addressing it now.

In reviewing the PR also found that this implementation can silently discard populated values. The null handling doesn't work consistently with `query_df()` or `use_none=False`, and local benchmarks showed ~4% slowdown for ordinary deeply nested JSON.

To address this issue fully we'd need an explicit collision policy, consistent behavior across all the query APIs and codecs, and no material overhead for unaffected reads before accepting a fix. We can revisit this if a concrete user workload demonstrates the need. Thanks again!

PR #1049, opened 2026-09-11
Fix SQLAlchemy 1.x CI dependency resolution

joe-clickhouse:
## Summary

The SQLAlchemy 1.x CI job installs Alembic 1.20, then downgrades SQLAlchemy alone. Alembic requires SQLAlchemy 2.x, so the resulting environment fails to import Alembic.

Resolve SQLAlchemy 1.x and Alembic together so pip selects a compatible pair, then run `pip check` before the tests.


joe-clickhouse:
This fix was cherry-picked into #1050 and merged there. Closing as superseded.

PR #1058, opened 2026-09-22
Make async external data replayable

### LINKED ISSUE #1057: Avoid blocking async external-data uploads while preserving retries
AI written, human reviewed.

### Describe the bug

The async HTTP backend passes byte-backed external-data file fields directly to `aiohttp.FormData.add_field()`. For large byte payloads, aiohttp warns that raw bytes are processed synchronously and may block the event loop.

Replacing the bytes with one file-like object while planning the request is not sufficient: retries can reuse a consumed or closed stream. The async backend constructs multipart form data inside its retry loop, so it can instead create a fresh `io.BytesIO` for each HTTP attempt.

### Steps to reproduce

1. Create an `ExternalData` table backed by a large `bytes` payload.
2. Submit it through `AsyncClient.query()`.
3. Observe aiohttp warning that passing raw bytes as a file field may block the event loop.
4. Retry the request or reuse the same `ExternalData` object and verify the complete payload must remain available.

### Expected behaviour

Byte-backed async multipart file fields should use a fresh file-like object for each request attempt. Large uploads should avoid aiohttp raw-byte blocking warnings, retries should receive the complete payload, and caller-owned `ExternalData` should remain reusable.

A minimal transport-level change is to wrap `bytes`, `bytearray`, and `memoryview` in `io.BytesIO` while constructing each multipart attempt:

```python
file_data = field_value[1]
if isinstance(file_data, (bytes, bytearray, memoryview)):
    file_data = io.BytesIO(file_data)
```

### Code example

```python
from clickhouse_connect.driver.external import ExternalData

payload = ExternalData(
    file_name="points.arrow",
    data=arrow_bytes,
    fmt="ArrowStream",
    structure="value UInt64",
)

first = await client.query("SELECT count() FROM points", external_data=payload)
second = await client.query("SELECT count() FROM points", external_data=payload)
```

The relevant aiohttp warning states that passing raw bytes as a file field is deprecated because it is processed synchronously and may block the event loop; a file-like object should be used instead.

#### Environment
* clickhouse-connect version: 1.8.0 and current `main`
* Python version: 3.13
* Operating system: Linux

#### ClickHouse server
* ClickHouse Server version: not server-version-specific
* ClickHouse Server non-default settings, if any: none
* `CREATE TABLE` statements for tables involved: none; query-scoped external table
* Sample data: Arrow-encoded `UInt64` values

TrapsterDK:
AI written, human reviewed.

## Summary
Fixes #1057.

The async HTTP backend now wraps byte-backed multipart file fields in a fresh `io.BytesIO` while constructing each request attempt. This avoids aiohttp's blocking raw-bytes file path while ensuring retries receive an unconsumed stream and caller-owned `ExternalData` remains reusable.

A regression test consumes and closes the first multipart stream, forces a stale-connection retry, and verifies the second attempt receives a distinct stream with the complete payload.

The test was confirmed to fail against the unpatched transport and pass with this change.

## Checklist


joe-clickhouse (approved):
Thanks @TrapsterDK.  I was able to reproduce the stalls during large TLS uploads. Your fix addresses the cause, and creating a fresh stream for each attempt preserves retries and reuse.

Note two corrections to the issue though. Retries already worked and the warning is a `ResourceWarning`, not a deprecation warning. The extra memory copy on plain HTTP is still a tradeoff but it's consistent with how the existing request-body handling works.

I kept your runtime fix unchanged. I merged main. On top of it, I replaced the mock-only test with a real integration test. I also corrected the changelog and moved the entry under UNRELEASED.

Thanks!

PR #1061, opened 2026-09-22
Add context manager protocol to DB-API Connection and Cursor

zatchet:
## Summary

The DB-API `Connection` and `Cursor` don't implement the context manager protocol, so the standard PEP 249 idioms fail:

```python
conn = clickhouse_connect.dbapi.connect(host="localhost")
with conn.cursor() as cursor:   # TypeError: 'Cursor' object does not support the context manager protocol
    cursor.execute("SELECT 1")

with conn:                       # TypeError: 'Connection' object does not support the context manager protocol
    ...
```

This adds `__enter__`/`__exit__` to both, closing on exit — the same shape as the existing `Client.__enter__`/`__exit__`.

One design note for review: some DB-API drivers scope a transaction on the connection context (sqlite3, psycopg2 commit-or-rollback on exit without closing). ClickHouse has no client-side transactions and this driver's `commit`/`rollback` are no-ops, so close-on-exit is the meaningful behavior here (as in duckdb or pyathena); happy to change if you prefer different semantics for the connection half — the cursor half is the uncontroversial part.

## Checklist

- Tests: cursor context closes on normal exit and on exception (subsequent fetches raise `ProgrammingError`), connection context returns the connection and calls `close` exactly once. `tests/integration_tests/test_dbapi.py` passes (20/20); the new tests fail on `main` with `TypeError`.
- `ruff check`/`ruff format` and `mypy` clean.
- `CHANGELOG.md` updated under `UNRELEASED`.


joe-clickhouse (approved):
Thanks @zatchet! I added the missing type annotations, exception-path test coverage, and clarified the docs/changelog to make the cleanup behavior explicit. Otherwise, looks good to me.

